PR #12, opened 2025-01-17
feat: engine driver

### LINKED ISSUE #5: Design and Implement Engine Driver
# Overview
We should implement a component that is used to drive the execution client via he Engine API.

Engine API reference: https://hackmd.io/@danielrachi/engine_api

greged93:
Add the engine driver component to the rollup node. The Engine driver is the interface to the Engine API of the EN and handles communicating new safe and unsafe payloads to the EN.

Resolves #5


crates/engine/src/payload.rs
```diff
@@ -0,0 +1,167 @@
...
+use alloy_primitives::{Bytes, B256};
+use alloy_rpc_types_engine::{ExecutionPayload, PayloadAttributes};
+use eyre::Result;
+use serde::{Deserialize, Serialize};
+use std::future::Future;
+use tracing::debug;
+
+/// The payload attributes for block building tailored for Scroll.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct ScrollPayloadAttributes {
```
greged93, line 11: I will move this to `scroll-alloy-rpc-types-engine` once the `alloy` PR is merge on Reth.

crates/engine/src/payload.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Indicates whether the payload building job should happen with or without pool transactions.
+    pub(crate) no_tx_pool: bool,
+}
+
+/// Returns true if the [`ScrollPayloadAttributes`] matches the [`ExecutionPayload`]:
+///    - provided parent hash matches the parent hash of the [`ExecutionPayload`]
+///    - all transactions match
+///    - timestamps are equal
+///    - `prev_randaos` are equal
+///    - TODO: should we also compare the `fee_recipient` with the `suggested_fee_recipient`?
```
greged93, line 25: they are compared in [OP](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/attributes/engine_consolidate.go#L64), but I think I remember @Thegaram mentioning that the `fee_recipient` field was used for something else currently

Thegaram: Is this `header.coinbase`? It is currently used by Clique.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+    ///   - If the payload is missing or doesn't match the attributes:
+    ///         - Starts payload building task on the EL via `engine_forkchoiceUpdatedV1`, passing
+    ///           the provided payload attributes.
+    ///         - Retrieve the payload with `engine_getPayloadV1`.
+    ///         - Sends the constructed payload to the EL via `engine_newPayloadV1`.
+    ///         - Sets the current fork choice for the EL via `engine_forkchoiceUpdatedV1`.
+    ///   - If the execution payload matches the attributes:
+    ///         - Sets the current fork choice for the EL via `engine_forkchoiceUpdatedV1`,
+    ///           advancing the safe head by one.
+    pub async fn handle_payload_attributes(
```
Thegaram, line 129: Will `handle_payload_attributes` be called by the derivation pipeline for new attributes?

crates/engine/src/payload.rs
```diff
@@ -0,0 +1,167 @@
...
+        return false
+    }
+
+    true
+}
+
+/// Implementers of the trait can provide the L2 execution payload for a block id.
+pub trait ExecutionPayloadProvider {
+    /// Returns the [`ExecutionPayload`] for the provided [`BlockId`], or [None].
+    fn execution_payload_by_block(
```
Thegaram, line 82: Can't this be simplified to `async fn`?

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+        let skip_attributes = maybe_execution_payload.as_ref().is_some_and(|ep| {
+            matching_payloads(&payload_attributes, ep, self.safe_block_info.hash)
+        });
+
+        if skip_attributes {
+            // set the safe head to the execution payload at safe block + 1 and issue forkchoice
+            self.set_safe_block_info(maybe_execution_payload.expect("exists").into());
+            self.forkchoice_updated(None).await?;
+        } else {
+            // start block building with `no_tx_pool = true`
```
Thegaram, line 146: I find it a bit counter-intuitive to call this _block building_, since the block body is already set, we just need to execute and recompute some header fields.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+            let id = self
+                .forkchoice_updated(Some(payload_attributes))
+                .await?
+                .ok_or_else(|| eyre!("missing payload id"))?;
+
+            // retrieve the execution payload
+            let execution_payload = self.get_payload(id).await?;
+
+            // issue the execution payload to the EL and set the new forkchoice
+            let safe_block_info = (&execution_payload).into();
```
Thegaram, line 157: I guess this is implicitly always true that this is `+ 1` compared to the current safe block height?

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.new_payload(execution_payload.into_v1()).await?;
+
+            self.set_safe_block_info(safe_block_info);
+            self.set_unsafe_block_info(safe_block_info);
```
Thegaram, line 161: While this should not happen during normal operation (we'll always get `skip_attributes` if the sequencer is non-equivocating and the derivation pipeline doesn't override the chain), it's interesting to consider what will happen if it does.

Because of the `forkchoice_updated` below our EN node will "snap" into the correct chain. So in this case any subsequent blocks gossiped by the malicious sequencer (that do not build on the `safe` block) will be ignored, i.e. we'll basically fall back to full L1 sync, is that correct?

Also a case that we can add to our test suite later.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+                .forkchoice_updated(Some(payload_attributes))
+                .await?
+                .ok_or_else(|| eyre!("missing payload id"))?;
+
+            // retrieve the execution payload
+            let execution_payload = self.get_payload(id).await?;
+
+            // issue the execution payload to the EL and set the new forkchoice
+            let safe_block_info = (&execution_payload).into();
+            self.new_payload(execution_payload.into_v1()).await?;
```
Thegaram, line 158: With "infallible chain derivation" we want to get to a point where this cannot error.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /// Calls `engine_newPayloadV1` and logs the result.
+    async fn new_payload(&self, execution_payload: ExecutionPayloadV1) -> Result<()> {
+        match self.client.new_payload_v1(execution_payload).await?.status {
+            PayloadStatusEnum::Invalid { validation_error } => {
+                error!(target: "engine::driver", ?validation_error, "failed to issue new execution payload");
+                bail!("invalid payload: {validation_error}")
+            }
+            PayloadStatusEnum::Syncing => {
```
Thegaram, line 175: Is it safe to continue in this case?

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+                debug!(target: "engine::driver", "EN syncing");
+            }
+            PayloadStatusEnum::Accepted => {
```
Thegaram, line 178: This should never happen in the `handle_payload_attributes` code path.

crates/engine/src/payload.rs
```diff
@@ -0,0 +1,167 @@
...
+        debug!(
+            target: "engine::driver",
+            expected = ?attributes.payload_attributes.timestamp,
+            got = ?payload.timestamp(),
+            "mismatch in timestamp"
+        );
+        return false
+    }
+
+    if payload.prev_randao() != attributes.payload_attributes.prev_randao {
```
Thegaram, line 66: Shouldn't we simply assert that `prev_randao` (and any other fields unused in Scroll) are zero/None?

Thegaram (approved):

crates/engine/src/payload.rs
greged93: It can, however adding `async fn` in a trait is syntaxing sugar for `fn my_func(...) -> impl Future<Output = ...>`, which means you can't specify marker traits on them (in this case `Send`). I think in this case, since we will be using `ExecutionPayloadProvider` as a generic, imo it is better to have it be send.
Ref: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/async_fn_in_trait/static.ASYNC_FN_IN_TRAIT.html

crates/engine/src/engine.rs
greged93: Yes indeed, any attempts to maliciously drive the chain into an incorrect state should be corrected by this step. However I just realized that I don't check that the execution payload received correctly builds on the previous `unsafe_head` so I will add this

crates/engine/src/engine.rs
greged93: I was actually missing an explicit ` self.set_unsafe_block_info(self.safe_block_info);` before the start of the payload building, I have added it now

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+        finalized_head: BlockInfo,
+    ) -> Self {
+        let fcu = ForkchoiceState {
+            head_block_hash: unsafe_head.hash,
+            safe_block_hash: safe_head.hash,
+            finalized_block_hash: finalized_head.hash,
+        };
+
+        // wait on engine
+        while let Err(err) = client.fork_choice_updated_v1(fcu, None).await {
```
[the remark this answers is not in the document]
greged93: will update to also trace the `Ok` path

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// The engine API client.
+    client: EC,
+    /// The execution payload provider
+    execution_payload_provider: P,
+    /// The unsafe L2 block info.
+    unsafe_block_info: BlockInfo,
+    /// The safe L2 block info.
+    safe_block_info: BlockInfo,
+    /// The finalized L2 block info.
+    finalized_block_info: BlockInfo,
```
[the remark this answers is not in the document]
greged93: yes makes sense!

crates/engine/src/engine.rs
greged93: I think it's fine, from the docs: 
```
 /// SYNCING is returned by the engine API in the following calls:
 ///   - newPayload:       if the payload was accepted on top of an active sync
```
So the payload is accepted and retained by the EN afaiu

crates/engine/src/engine.rs
greged93: you mean the syncing status? I'm not sure why this should never happen in `handle_payload_attributes`, could you elaborate?

crates/engine/src/payload.rs
greged93: I can add an assert on the fields that we expect to be zero or None yes, but this should probably go into another function no? Here we are just checking for a match between the execution payload and the attributes. I can add a validity check of the attributes/execution payload

crates/engine/src/engine.rs
greged93: I don't think it would be the responsibility of the engine driver to ensure this. We should have a higher abstraction which handles the caching of derived attributes and can also retry calling `handle_payload_attributes` if we error, similar to how Kona will have an attributes queue.

crates/engine/src/engine.rs
greged93: > The derivation pipeline will always provide each block

I think that would make syncing extremely slow no? Afaiu [Optimism](https://specs.optimism.io/protocol/exec-engine.html?highlight=fast%20sync#sync) does allow the EN to sync via its p2p, meaning you could enter the `Sync` status. I need to clarify how you enter this "Happy-path" for OP.

crates/engine/src/engine.rs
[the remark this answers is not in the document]
Thegaram: True, I missed that, my bad

crates/engine/src/engine.rs
frisitano: The `EngineDriver` would not be a component of the provable derivation pipeline. Instead, we would have an `EvmExecutor` component that uses `revm` to execute the block in the same process so that it can be proven. We may want to introduce a trait such that the `EngineDriver` can be easily replaced by the `EvmExecutor` for the provable derivation pipeline however we should probably explore that in future PR's / milestones.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+                .forkchoice_updated(Some(payload_attributes))
+                .await?
+                .ok_or_else(|| eyre!("missing payload id"))?;
+
+            // retrieve the execution payload
+            let execution_payload = self.get_payload(id).await?;
+
+            // issue the execution payload to the EL and set the new forkchoice
+            let safe_block_info = (&execution_payload).into();
+            self.new_payload(execution_payload.into_v1()).await?;
```
frisitano, line 171: Do we need to invoke `new_payload` after the EL has just provided us the payload?

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+                error!(target: "engine::driver", ?validation_error, "failed to issue new execution payload");
+                bail!("invalid payload: {validation_error}")
+            }
+            PayloadStatusEnum::Syncing => {
+                debug!(target: "engine::driver", "EN syncing");
+            }
+            PayloadStatusEnum::Accepted => {
+                warn!(target: "engine::driver", "execution payload part of side chain");
+            }
+            _ => {}
```
frisitano, line 194: Should we add a `trace!` log for the `Valid` case?

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+                error!(target: "engine::driver", ?validation_error, "failed to issue forkchoice");
+                bail!("invalid fork choice: {validation_error}")
+            }
+            PayloadStatusEnum::Syncing => {
+                debug!(target: "engine::driver", "EN syncing");
+            }
+            PayloadStatusEnum::Accepted => {
+                warn!(target: "engine::driver", "payload attributes part of side chain");
+            }
+            _ => {}
```
frisitano, line 219: Let's add a `trace!` log for the valid case.

crates/engine/src/payload.rs
```diff
@@ -0,0 +1,167 @@
...
+            got = ?payload.parent_hash(),
+            "mismatch in parent hash"
+        );
+        return false
+    }
+
+    let payload_transactions = &payload.as_v1().transactions;
+    let matching_transactions = payload_transactions.len() ==
+        attributes.transactions.as_ref().map(|v| v.len()).unwrap_or_default() &&
+        attributes.transactions.as_ref().is_some_and(|v| v == payload_transactions);
```
frisitano, line 44: Doesn't rust also check the length of vectors during comparison?

```suggestion
    let payload_transactions = &payload.as_v1().transactions;
    let matching_transactions =
        attributes.transactions.as_ref().is_some_and(|txs| txs == payload_transactions);
```

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+use tracing::{debug, error, info, warn};
+
+/// The main interface to the Engine API of the EN.
+/// Internally maintains the fork state of the chain.
+#[derive(Debug, Clone)]
+pub struct EngineDriver<EC, P, ET> {
+    /// The engine API client.
+    client: EC,
+    /// The execution payload provider
+    execution_payload_provider: P,
```
frisitano, line 22: How are you thinking about the `client` and `execution_payload_provider`? Wouldn't they both be the same `alloy` provider client?

frisitano (commented):
Looks good! Added a few comments inline.

crates/engine/src/engine.rs
Thegaram: I think it will be structured this way:
- We will have a `pipeline` crate that implements the derivation pipeline, except for EVM execution (so derive PayloadAttributes, not blocks). The adjective _provable_ involves multiple things: It probably needs to be `no_std` and it should only operate on data pulled from controlled sources (e.g. `BlobDataSource`, `BatchDataSource` traits) that we can verify during the proving process.
- We will also have a `scroll-revm` crate (already done).

We will use these crates in two main settings:
- **Scroll node**: `pipeline` runs in RN, pushes payloads through `EngineDriver` into EN into `scroll-revm` for execution.
- **Scroll prover**: Stadalone binary that uses both `prover` and `scroll-revm`, it runs on a zkVM. (See `op-succinct` for an example: [derive](https://github.com/succinctlabs/op-succinct/blob/dc6bbb5a14e5d5c07bff2d7b358c63e16fa2c65c/programs/range/src/main.rs#L252) and [prove](https://github.com/succinctlabs/op-succinct/blob/dc6bbb5a14e5d5c07bff2d7b358c63e16fa2c65c/programs/range/src/main.rs#L352).)

Thegaram: My specific feedback here was:

The derivation pipeline feeds blocks in a certain way: Continuously each block one-by-one, or fork off from a previous block (L1 reorg). But there will never be gaps.

As a result, when we call the EngineAPI in `handle_payload_attributes`, it should certainly never return `SYNCING`, and maybe also not `INVALID` and `ACCEPTED`.

It's fine if we just log and observe these for now. But later on I'd prefer adding some more sanity checks here.

crates/engine/src/engine.rs
Thegaram: This should never happen for the derivation pipeline. But if it does, we should definitely not keep pushing new blocks, we need to first make sure that previous blocks were not processed as `INVALID` (otherwise the whole L1 sync will get stuck).

No need to change now, just let's keep in mind for the future. Feel free to resolve.

crates/engine/src/engine.rs
Thegaram: All statuses are possible in `handle_execution_payload` (p2p sync).

[Not all statuses are possible](https://github.com/scroll-tech/rollup-node/pull/12#discussion_r1922552467) in `handle_execution_payload` (L1 sync).

Later on it might be useful to differentiate between these two code paths. For now feel free to resolve.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+        execution_payload: ExecutionPayload,
+    ) -> Result<()> {
+        let unsafe_block_info = (&execution_payload).into();
+        let execution_payload = execution_payload.into_v1();
+        self.new_payload(execution_payload).await?;
+        self.set_unsafe_block_info(unsafe_block_info);
+        self.forkchoice_updated(None).await?;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
Thegaram: Afaik the sequencer mode (payload building) is not implemented in this PR. We only have:
- `handle_execution_payload`: execute block received in L2 p2p
- `handle_payload_attributes`: execute block derived from L1 data

crates/engine/src/engine.rs
Thegaram: `execution_payload_provider` could be implemented using a regular Ethereum JSON-RPC client (not Engine API client). But maybe they could be the same, too. I think it's nicer to decouple them here.

greged93: I think there is a default implementation for the client via the proc macro which derives this:
```rust
impl<TypeJsonRpseeInteral, Engine: EngineTypes> EngineApiClient<Engine> for TypeJsonRpseeInteral where TypeJsonRpseeInteral: jsonrpsee::core::client::ClientT, Engine::PayloadAttributes: jsonrpsee::core::Serialize + Clone {}
``` 
I had the `client` and the `execution_payload_provider` separated because we don't necessarily need to always fetch the execution payload from the EN, I think [optimism](https://github.com/ethereum-optimism/optimism/blob/39e9f1a4693912af8bdef428d246c15a6cf44ec7/op-program/client/l2/engine_backend.go#L23) caches the execution payload received from the P2P layer in memory. At least this is what I understood from [here](https://github.com/ethereum-optimism/optimism/blob/49ea3d0be3a5a7e2c16fa6a5cffb6c29ab3dcbab/op-node/rollup/attributes/attributes.go#L171).

crates/engine/src/engine.rs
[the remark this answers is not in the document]
frisitano: It makes sense to leave sequencer-specific requirements to a follow-up PR. Technically, the `handle_payload_attributes` method uses EN payload building to compute the block from `ScrollPayloadAttributes`.

crates/engine/src/engine.rs
frisitano: The `ScrollPayloadAttributes` would come from the L1 (via the derivation pipeline), but the `ExecutionPayload` comes from the EN via a payload-building job. The payload-building job is requested above on line 161 via the `forkchoice_updated` invocation, and then the result (`ExecutionPayload`) is retrieved on line 167 (`get_payload`). I'm unsure if the engine will automatically add the built payload to the chain tree or if the `new_payload` invocation is required.

crates/engine/src/engine.rs
frisitano: I think it would be cleaner to have a single generic client that is bound by `EngineApiClient` and `ExecutionPayloadProvider` however let's leave it as is for now and consider this in a follow-up PR.

crates/engine/src/engine.rs
greged93: I don't think `forkchoice_updated` inserts the payload in the tree, it starts the building process, returning the id for it but I can't find anything in the `EngineTreeApiHandler` that suggests it also appends it to the chain.
Also I added this because this is the flow followed by [OP](https://arc.net/l/quote/zqqvdfkq).

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+{
+    /// Initialize the driver and wait for the Engine server to be ready.
+    pub async fn init_and_wait_for_engine(
+        client: EC,
+        execution_payload_provider: P,
+        unsafe_head: BlockInfo,
+        safe_head: BlockInfo,
+        finalized_head: BlockInfo,
+    ) -> Self {
+        let fcu = ForkchoiceState {
```
georgehao, line 50:
```suggestion
        let fcs = ForkchoiceState {
```

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+                Err(err) => {
+                    debug!(target: "engine::driver", ?err, "waiting on engine client");
+                    tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+                }
+                Ok(status) => {
+                    info!(target: "engine::driver", payload_status = ?status.payload_status.status, "engine ready");
+                    break
+                }
+            }
+        }
```
georgehao, line 68:
```suggestion
                // wait on engine
        while let Err(err) = client.fork_choice_updated_v1(fcu, None).await {
            debug!(target: "engine::driver", ?err, "waiting on engine client");
            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
        }

        info!(target: "engine::driver", "engine ready");
```

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// The execution payload provider
+    execution_payload_provider: P,
+    /// The unsafe L2 block info.
+    unsafe_block_info: BlockInfo,
+    /// The safe L2 block info.
+    safe_block_info: BlockInfo,
+    /// The finalized L2 block info.
+    finalized_block_info: BlockInfo,
+    /// Marker
+    _types: std::marker::PhantomData<ET>,
```
georgehao, line 30: just curious about why this variable need a `_`

crates/engine/src/engine.rs
Thegaram: We need to keep `Ok(status)` to log it

crates/engine/src/engine.rs
Thegaram: `PhantomData` is a zero-sized type that is only there to give a hint about `ET` to the compiler. So this field will never be accessed. The `_` prefix just tells the compiler that this field is not used and prevents warnings (see [here](https://www.reddit.com/r/rust/comments/ynuf0q/what_is_expressions_in_rust/)).

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+            finalized_block_info: finalized_head,
+            _types: std::marker::PhantomData,
+        }
+    }
+
+    /// Reorgs the driver by setting the safe and unsafe block info to the finalized info.
+    pub fn reorg(&mut self) {
+        self.unsafe_block_info = self.finalized_block_info;
+        self.safe_block_info = self.finalized_block_info;
+    }
```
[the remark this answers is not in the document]
greged93: yes sorry, forgot to update this in the issue which caused the incorrect implementation here

crates/engine/src/engine.rs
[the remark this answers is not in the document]
greged93: > Technically, the handle_payload_attributes method uses EN payload building to compute the block from ScrollPayloadAttributes.

Yes I don't think there is much to had for allow for block building, correct me if I am wrong but I think we need to:
- have `handle_payload_attributes` return the execution payload instead of inserting it, in order to gossip the block as soon as possible.
- pass the payload to the EN via `handle_execution_payload`

crates/engine/src/payload.rs
greged93: Yes you are right

crates/engine/src/engine.rs
frisitano: > We will use these crates in two main settings:
> 
> * **Scroll node**: `pipeline` runs in RN, pushes payloads through `EngineDriver` into EN into `scroll-revm` for execution.
> * **Scroll prover**: Stadalone binary that uses both `prover` and `scroll-revm`, it runs on a zkVM. (See `op-succinct` for an example: [derive](https://github.com/succinctlabs/op-succinct/blob/dc6bbb5a14e5d5c07bff2d7b358c63e16fa2c65c/programs/range/src/main.rs#L252) and [prove](https://github.com/succinctlabs/op-succinct/blob/dc6bbb5a14e5d5c07bff2d7b358c63e16fa2c65c/programs/range/src/main.rs#L352).)

Yes, this is also how I'm thinking about it.

> It's fine if we just log and observe these for now. But later on I'd prefer adding some more sanity checks here.

Agreed, let's add a `TODO` in the code.

crates/engine/src/engine.rs
frisitano: Got it, thanks for the reference and explanation that helps. So we can consider payload building and consensus as separate entities/tasks.

crates/engine/src/engine.rs
```diff
@@ -0,0 +1,245 @@
...
+            finalized_block_info: finalized_head,
+            _types: std::marker::PhantomData,
+        }
+    }
+
+    /// Reorgs the driver by setting the safe and unsafe block info to the provided fork info.
+    pub fn reorg(&mut self, fork: BlockInfo) {
+        self.unsafe_block_info = fork;
+        self.safe_block_info = fork;
+    }
```
frisitano, line 87: Is this the correct logic? Say we have an unsafe chain reorg but the safe tip remains the same we would want to exclusively fork the unsafe chain but not the safe one. I would suggest that we remove this for now and reintroduce it when we need it.

frisitano (approved):
Looks good! Left a single comment inline. Let me know what you think.

crates/engine/src/engine.rs
greged93: yes I was also thinking about leaving this reorg logic handling for a higher level component (watchers or manager), will remove

PR #17, opened 2025-01-24
Bridge Implementation

frisitano:
# Overview
This PR introduces a bridge between the `eth-wire` and `scroll-wire` protocol. This is done by implementing a `BridgeBlockImport` which implements `BlockImport` and is used in the "inner network manager". The `BridgeBlockImport` type wraps an inner block import type. The `BridgeBlockImport` delegates importing of blocks to the `inner_block_import`.  Upon a Valid block import outcome it attempts to extract the signature from the `extra_data` field of the block and upon success sends the block to the scroll `NetworkManager`. Changes to other components of the stack have also been implemented to support this approach. Some general code hygiene changes have also been implemented.


bin/bridge/src/import.rs
```diff
@@ -0,0 +1,108 @@
...
+        // extra data field.
+        if let Some(signature) = extra_data
+            .len()
+            .checked_sub(ECDSA_SIGNATURE_LEN)
+            .and_then(|i| Signature::from_compact(&extra_data[i..]).ok())
+        {
+            let block = block.block.clone();
+            trace!(target: "bridge::import", peer_id = %peer_id, block = ?block, "Received new block from eth-wire protocol");
+
+            // We trigger a new block event to be sent to the network manager. If this results in an
```
greged93, line 60:
```suggestion
            // We trigger a new block event to be sent to the rollup node's network manager. If this results in an
```

crates/network/src/manager.rs
```diff
@@ -4,81 +4,127 @@ use super::{
...
+///   used to bridge new blocks announced on the eth-wire protocol to the scroll-wire protocol.
 pub struct NetworkManager {
     /// A handle to the inner reth network manager.
     inner_network_handle: RethNetworkHandle,
     /// Handles block imports for new blocks received from the network.
     block_import: Box<dyn BlockImport>,
-    /// The receiver half of the channel set up between this type and the [`NetworkHandle`], receives
-    /// commands from the [`NetworkHandle`].
+    /// The receiver half of the channel set up between this type and the [`NetworkHandle`],
+    /// receives commands from the [`NetworkHandle`].
```
greged93, line 41:
```suggestion
    /// The sender half of the channel set up between this type and the [`NetworkHandle`],
    /// sends commands to the [`NetworkHandle`].
```

crates/network/src/manager.rs
```diff
@@ -4,81 +4,127 @@ use super::{
...
     to_manager_tx: UnboundedSender<NetworkHandleMessage>,
     /// Receiver half of the channel set up between this type and the [`NetworkHandle`], receives
     /// [`NetworkHandleMessage`]s.
     from_handle_rx: UnboundedReceiverStream<NetworkHandleMessage>,
```
greged93, line 45: we should probably stay consistent here and have `to_handle_tx` or `from_manager_rx`.

crates/scroll-wire/src/manager.rs
```diff
@@ -1,103 +1,86 @@
...
     /// Announces a new block to the specified peer.
-    pub fn announce_block(&mut self, peer_id: PeerId, block: NewBlock, hash: B256) {
+    pub fn announce_block(&mut self, peer_id: PeerId, block: &NewBlock, hash: B256) {
         if let Entry::Occupied(to_connection) = self.connections.entry(peer_id) {
-            // if the channel to the peer is closed then we can remove the connection and state from the manager.
-            let connection = to_connection.get();
-            if connection.is_closed() {
+            // We send the block to the peer. If we receive an error we remove the peer from the
+            // connections map and delete its state as the connection is no longer valid.
+            if to_connection.get().send(Message::new_block(block.clone())).is_err() {
```
greged93, line 43: would we want to log the error here?

greged93 (commented):
Couple of nits and questions but looks good!

bin/bridge/src/main.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+#[cfg(any(feature = "optimism", not(feature = "scroll")))]
+fn main() {
+    eprintln!("Scroll feature is not enabled");
+    std::process::exit(1);
+}
+
+#[cfg(test)]
+mod test {
```
greged93, line 63: this looks like an integration type of test, we could move it to a `tests` folder

bin/bridge/src/main.rs
```diff
@@ -0,0 +1,322 @@
...
+                )
+                .with_add_ons(ScrollAddOns::default())
+                .launch_with_fn(|builder| {
+                    let launcher = EngineNodeLauncher::new(
+                        builder.task_executor().clone(),
+                        builder.config().datadir(),
+                        engine_tree_config,
+                    );
+                    builder.launch_with(launcher)
+                })
```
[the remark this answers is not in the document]
frisitano: it appears to be compiling fine.

bin/bridge/src/main.rs
```diff
@@ -0,0 +1,322 @@
...
+use network::ScrollBridgeNetworkBuilder;
+
+mod import;
+use import::BridgeBlockImport;
+
+#[global_allocator]
+static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
+
+#[cfg(all(feature = "scroll", not(feature = "optimism")))]
+fn main() {
```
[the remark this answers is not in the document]
frisitano: Agreed. Let's keep it in mind and see if more demand for this code it surfaces.

crates/network/src/manager.rs
```diff
@@ -116,7 +160,6 @@ impl NetworkManager {
     fn on_handle_message(&mut self, message: NetworkHandleMessage) {
         match message {
             NetworkHandleMessage::AnnounceBlock { block, signature } => {
```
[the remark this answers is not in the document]
frisitano: The idea would be that the rollup node manager stores a handle to the network manager. Upon the success of both consensus and validity checks (orchestrated by the rollup node manager) on the block we would announce the block to the network via the network handle.

crates/network/src/manager.rs
frisitano: `to_handle_tx` and `from_manager_rx` don't make sense in this context as there is a one-way message channel from the handle to the manager. So "`to_handle_tx` = transmit to handle" isn't possible and equal argument for `from_manager_rx`. This is my interpretation, what do you think?

crates/scroll-wire/src/manager.rs
frisitano: The error means that the channel has been closed but does not provide additional context, it is a tokio `SendError`.

crates/network/src/manager.rs
greged93: Ahhhh ok I get what you mean, yes let's leave as it is then

bin/bridge/src/network.rs
```diff
@@ -0,0 +1,85 @@
...
+        // Create a scroll-wire protocol handler.
+        let (scroll_wire_handler, events) = ProtocolHandler::new(ScrollWireConfig::new(true));
+
+        // Create the network configuration.
+        let config = ctx.network_config()?;
+        let mut config = NetworkConfig {
+            network_mode: NetworkMode::Work,
+            block_import: Box::new(super::BridgeBlockImport::new(
+                new_block_tx,
+                self.block_import.unwrap_or_else(|| config.block_import),
```
frisitano, line 62: The default block import is the `ProofOfStakeBlockImport` which is a noop and as such this will result in no blocks being propagate. I propose that initially we replace this with the `ValidBlockImport` and later move to a method that validates blocks via the engine API prior to propagating.

greged93: that works for me

bin/bridge/src/main.rs
```diff
@@ -4,6 +4,9 @@ use network::ScrollBridgeNetworkBuilder;
 mod import;
 use import::BridgeBlockImport;
 
+#[cfg(test)]
```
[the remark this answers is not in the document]
frisitano: updated. I've added a `test-utils` feature which we will need to run the integration test. This may throw some errors when we look to integrate CI in your PR.

bin/bridge/src/import.rs
```diff
@@ -59,9 +59,10 @@ impl BridgeBlockImport {
             let block = block.block.clone();
             trace!(target: "bridge::import", peer_id = %peer_id, block = ?block, "Received new block from eth-wire protocol");
 
-            // We trigger a new block event to be sent to the network manager. If this results in an
-            // error it means the network manager has been dropped.
-            let _ = self.events.send(Event::NewBlock { peer_id, block, signature });
+            // We trigger a new block event to be sent to the rollup node's network manager. If this
+            // results in an
```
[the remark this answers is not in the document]
frisitano: good catch.

greged93 (approved):
lgtm!

PR #18, opened 2025-01-31
feat: ci

### LINKED ISSUE #3: Design and Implement CI
# Overview
We should design and implement our CI pipeline.

greged93:
- Adds required CI checks to the repo.
- Fixes lints, formatting, docs and imports issues.

Resolves #3


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
greged93, line 1: Latest recommendation for the Rust ecosystem is to add the `Cargo.lock` file to the version control, and either use dependabot or `cargo update && cargo test` in CI.
https://github.com/rust-lang/cargo/issues/8728

.github/workflows/test.yaml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          toolchain: 'nightly'
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+      - uses: taiki-e/install-action@nextest
+      - name: Run unit tests
+        run: cargo nextest run --workspace --locked -E '!kind(test)'
+
+  it:
```
frisitano, line 34: maybe it would be preferable to be more verbose and name this `integration`.

frisitano (commented):
Looks good! One proposal - should we include a `Makefile` that allows us to run the lints locally? This will allow us to assert compliance with CI before pushing and could speed up dev cycles.

greged93:
Agree, on it!

frisitano (approved):
Looks great, thanks for this! LGTM

PR #20, opened 2025-02-04
fix: service restart

greged93:
Avoids the shutdown of the service on multiple restart burst in a short interval. With a restart of 1 second, 5 consecutive restarts in a short interval (< 10 seconds) will trigger the default [`StartLimitBurst` option](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#StartLimitIntervalSec=interval), permanently shutting down the service.


frisitano (approved):
LGTM

PR #21, opened 2025-02-07
Rollup node manager

frisitano:


crates/network/src/manager.rs
```diff
@@ -33,21 +36,15 @@ use tracing::trace;
...
+#[derive(Debug)]
 pub struct NetworkManager {
-    /// A handle to the inner reth network manager.
-    inner_network_handle: RethNetworkHandle<ScrollNetworkPrimitives>,
-    /// Handles block imports for new blocks received from the network.
-    block_import: Box<dyn BlockImport>,
-    /// The sender half of the channel set up between this type and the [`NetworkHandle`],
-    /// sends commands to the [`NetworkHandle`].
-    to_manager_tx: UnboundedSender<NetworkHandleMessage>,
+    /// A handle used to interact with the network manager.
```
frisitano, line 41: Fix doc comment on this type after updating the fields.

crates/node/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+pub struct RollupNodeManager<C, EC, P> {
+    /// The network handle used to communicate with the network manager.
+    network: NetworkManager,
+    engine: Arc<EngineDriver<EC, P>>,
+    consensus: C,
+    new_block_rx: UnboundedReceiverStream<NewBlockWithPeer>,
+    forkchoice_state: ForkchoiceState,
+    pending_block_imports:
+        FuturesUnordered<Pin<Box<dyn Future<Output = BlockImportOutcome> + Send>>>,
+}
```
frisitano, line 57: Review comments

crates/node/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+use tokio::sync::mpsc::UnboundedReceiver;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+use tracing::{error, trace};
+
+mod consensus;
+use consensus::Consensus;
+pub use consensus::PoAConsensus;
+
+// mod error;
+// use error::EngineManagerError;
```
frisitano, line 33: Purge

crates/node/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+            ExecutionPayloadV1::from_block_slow(&block).into();
+        let unsafe_block_info: BlockInfo = (&execution_payload).into();
+        self.forkchoice_state.update_unsafe_block_info(unsafe_block_info);
+        let fcs = self.get_fcs();
+
+        // Send the block to the engine to validate the correctness of the block.
+        let engine = self.engine.clone();
+        let future = Box::pin(async move {
+            trace!(target: "scroll_rollup_manager::RollupNodeManager", "handling block import future");
+            let result = match engine.handle_execution_payload(execution_payload, fcs).await {
```
frisitano, line 114: Review error handling here.

crates/node/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+    C: Consensus + Unpin,
+    EC: ScrollEngineApi<ScrollNetwork> + Unpin + Sync + Send + 'static,
+    P: ExecutionPayloadProvider + Unpin + Send + Sync + 'static,
+{
+    type Output = ();
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+
+        trace!("Polling RollupNodeManager");
```
frisitano, line 167: remove

crates/scroll-wire/Cargo.toml
```diff
@@ -17,14 +17,16 @@ alloy-primitives = { workspace = true, features = [
...
+reth-eth-wire = { git = "https://github.com/scroll-tech/reth.git", branch = "feat/add-deref-blanket-engine-api" }
 reth-network.workspace = true
 reth-network-api.workspace = true
+reth-primitives.workspace = true
 
 # scroll
 reth-scroll-primitives = { workspace = true, features = ["serde"] }
 
 # misc
+eyre.workspace = true
```
frisitano, line 29: Do we need this?

crates/scroll-wire/src/protocol/handler.rs
```diff
@@ -1,26 +1,26 @@
...
+use super::ScrollWireEvent;
+use crate::{connection::ScrollConnectionHandler, ScrollWireConfig};
 use reth_network::protocol::ProtocolHandler as ProtocolHandlerTrait;
 use reth_network_api::PeerId;
 use tokio::sync::mpsc;
 
-/// A Receiver for [`Event`].
-pub(super) type ScrollWireEventReceiver = mpsc::UnboundedReceiver<Event>;
+/// A Receiver for `ScrollWireEvents`.
+pub(super) type ScrollWireEventReceiver = mpsc::UnboundedReceiver<ScrollWireEvent>;
```
frisitano, line 8: Is this type needed?

crates/scroll-wire/src/protocol/proto.rs
```diff
@@ -45,54 +45,54 @@ impl TryFrom<u8> for MessageId {
...
+    pub id: ScrollMessageId,
+    pub payload: ScrollMessagePayload,
 }
 
-impl Message {
-    /// Returns the capability of the `ScrollWire` protocol.
+impl ScrollMessage {
+    /// Returns the capability of the scroll wire protocol.
     pub const fn capability() -> Capability {
         Capability::new_static("scroll-wire", 1)
```
frisitano, line 58: This could be changed to `scroll` instead.

frisitano (commented):
Added comments inline.

crates/engine/src/engine.rs
```diff
@@ -70,124 +57,105 @@ where
...
+            .client
+            .new_payload_v1(execution_payload)
+            .await
+            .map_err(|_| EngineDriverError::EngineUnavailable)?
+            .status
+        {
             PayloadStatusEnum::Invalid { validation_error } => {
                 error!(target: "engine::driver", ?validation_error, "failed to issue new execution payload");
-                bail!("invalid payload: {validation_error}")
+                return Err(EngineDriverError::InvalidExecutionPayload)
```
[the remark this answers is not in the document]
frisitano: We want to exit early and not invoke a FCU in the invalid case and as such we should return an error.

bin/bridge/src/network.rs
```diff
@@ -72,12 +60,37 @@ where
...
+        let engine = EngineDriver::new(engine_api, payload_provider);
+
+        let rollup_node_manager = RollupNodeManager::new(
+            scroll_network_manager,
+            engine,
+            ForkchoiceState::new(
+                BlockInfo { number: 0, hash: Default::default() },
+                BlockInfo { number: 0, hash: Default::default() },
+                BlockInfo { number: 0, hash: Default::default() },
+            ),
```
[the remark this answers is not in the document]
frisitano: Instead we implement `pub const fn genesis(chain: NamedChain) -> Self`

crates/node/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+        while let Poll::Ready(Some((block_info, outcome))) =
+            this.pending_block_imports.poll_next_unpin(cx)
+        {
+            this.handle_block_import_outcome(block_info, outcome);
+        }
+
+        // Handle blocks received from the eth-wire protocol.
+        while let Poll::Ready(Some(block)) = this.new_block_rx.poll_next_unpin(cx) {
+            this.handle_new_block(block);
+            cx.waker().wake_by_ref();
```
greged93, line 234: what is the reason for calling the waker here? Is it because of this sentence in the FuturesUnordered docs?

> When new futures are added, [poll_next](https://docs.rs/futures/latest/futures/prelude/trait.Stream.html#tymethod.poll_next) must be called in order to begin receiving wake-ups for new futures.

greged93 (commented):
Looks good, I think the refactor of the Networking make the code flow clearer! Just couple of nits and questions

crates/engine/src/engine.rs
```diff
@@ -1,239 +1,237 @@
...
+    ) -> Result<(PayloadStatusEnum, PayloadStatusEnum), EngineDriverError> {
+        // Convert the payload to the V1 format.
         let execution_payload = execution_payload.into_v1();
-        self.new_payload(execution_payload).await?;
-        self.set_unsafe_block_info(unsafe_block_info);
-        self.forkchoice_updated(None).await?;
 
-        Ok(())
+        // Invoke the FCU with the new state.
+        let fcu = self.forkchoice_updated(fcs, None).await?;
```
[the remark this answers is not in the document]
frisitano: This function receives the latest FCS from the rollup node manager as an argument. The rollup node manager will determine this fork choice state based on the blocks received from L2 P2P and L1 blocks from the `L1Observer`. To do so, the rollup node manager must be aware of L1 and L2 reorgs.

crates/engine/src/engine.rs
```diff
@@ -1,61 +1,48 @@
...
     /// The execution payload provider
     execution_payload_provider: P,
-    /// The unsafe L2 block info.
-    unsafe_block_info: BlockInfo,
-    /// The safe L2 block info.
-    safe_block_info: BlockInfo,
-    /// The finalized L2 block info.
-    finalized_block_info: BlockInfo,
-    /// Marker
-    _types: std::marker::PhantomData<ET>,
```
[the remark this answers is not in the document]
frisitano: It does not mutate `client` and `execution_payload_provider` hence a lock is not required.

crates/engine/src/engine.rs
```diff
@@ -1,239 +1,237 @@
...
-            let id = self
-                .forkchoice_updated(Some(payload_attributes))
-                .await?
-                .ok_or_else(|| eyre!("missing payload id"))?;
+            let fc_updated = self.forkchoice_updated(fcs, Some(payload_attributes)).await?;
 
             // retrieve the execution payload
-            let execution_payload = self.get_payload(id).await?;
+            let execution_payload = self
+                .get_payload(fc_updated.payload_id.expect("payload attributes has been set"))
```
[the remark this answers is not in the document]
frisitano: Given that we initiate a payload building job on line 138 we would expect `fc_updated.payload_id` to be `Some(..)` as per the EngineAPI spec. Do you think we should be handling error cases where the EngineAPI doesn't conform to spec?

crates/node/src/lib.rs
frisitano: yes exactly, good catch! I've updated the code to ensure we do this for all code paths where a new pending import is added to the data structure.

crates/network/src/handle.rs
```diff
@@ -50,6 +50,10 @@ impl NetworkHandle {
         let _ = self.inner.to_manager_tx.send(msg);
     }
 
+    pub fn block_import_outcome(&self, outcome: super::BlockImportOutcome) {
+        self.send_message(NetworkHandleMessage::BlockImportOutcome(outcome));
+    }
+
     /// Announces a block to the network.
     pub fn announce_block(&self, block: ScrollBlock, signature: Signature) {
```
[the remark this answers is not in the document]
frisitano: Currently, it's not used but we will use it when we have the sequencer implemented to gossip blocks to the network. I think it's fine to keep, as we know it will be used in the future. However, I'm not strongly opinionated and happy to remove it if you think it's best.

crates/network/src/manager.rs
```diff
@@ -139,17 +115,17 @@ impl NetworkManager {
...
         for peer_id in peers {
-            trace!(target: "network::manager", peer_id = %peer_id, block_hash = %hash, "Announcing new block to peer");
+            trace!(target: "scroll::network::manager", peer_id = %peer_id, block_hash = %hash, "Announcing new block to peer");
             self.scroll_wire.announce_block(peer_id, &block, hash);
         }
     }
 
     /// Handler for received events from the [`ScrollWireManager`].
-    fn on_scroll_wire_event(&mut self, event: Event) {
+    fn on_scroll_wire_event(&mut self, event: ScrollWireEvent) -> Option<NetworkManagerEvent> {
```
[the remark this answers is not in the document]
frisitano: My thinking was that in the future we would have other `ScrollWireEvent`s and some of them would not result in a `NetworkManagerEvent`. However, given the current implementation, I agree. We can always revert in the future if needed.

crates/engine/src/engine.rs
```diff
@@ -1,239 +1,235 @@
...
-        self.new_payload(execution_payload).await?;
-        self.set_unsafe_block_info(unsafe_block_info);
-        self.forkchoice_updated(None).await?;
 
-        Ok(())
+        // Issue the new payload to the EN.
+        let payload_status = self.new_payload(execution_payload).await?;
+
+        // Invoke the FCU with the new state.
+        let fcu = self.forkchoice_updated(fcs, None).await?;
```
greged93, line 84: should we do `fcs.safe_block_hash = unsafe_block_info.hash;` before issuing the forkchoice?

crates/engine/src/engine.rs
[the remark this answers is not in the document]
greged93: I'm maybe overcautious on this, it's true that it should always be some given the way we issue the fc

crates/network/src/handle.rs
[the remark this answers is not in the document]
greged93: let's keep it

greged93 (approved):
just a remark on the engine, otherwise lgtm!

crates/engine/src/engine.rs
frisitano: The logic below defines what is currently implemented to manage fcs. We get the current fcs from the node manager and then set the head block hash to the block hash of the unsafe block received from the network when defining the future for importing the block. Do you think we should be setting the safe block hash when we receive a new block over p2p?
https://github.com/scroll-tech/rollup-node/blob/04a5f4444479add9d45f1317fa81219c72feecba/crates/node/src/lib.rs#L145-L178

Upon block import result if success we update the fcs of the node manager.
https://github.com/scroll-tech/rollup-node/blob/04a5f4444479add9d45f1317fa81219c72feecba/crates/node/src/lib.rs#L197-L214

This works ok but I'm thinking we should introduce a `State` type that can be shared across async tasks (`type State = Arc<Mutex<InnerState>>`

PR #32, opened 2025-03-12
feat: watcher

### LINKED ISSUE #4: Design Watcher Component
# Overview
The watcher component is responsible for interacting with the L1 and L2 nodes via RPC.

greged93:
Introduces the L1 watcher in the rollup node.

Resolves #4


crates/primitives/src/batch.rs
```diff
@@ -0,0 +1,148 @@
+use alloy_primitives::{BlockNumber, B256};
```
greged93, line 1: This should be removed based on the decision from the primitives PR and the Pipeline interface.

crates/watcher/src/constants.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// The address of the Scroll Rollup contract on the L1.
+pub const ROLLUP_CONTRACT_ADDRESS: Address = address!("0xa13BAF47339d63B743e7Da8741db5456DAc1E556");
+
+/// The address of the Scroll L1 message queue contract on the L1.
+pub const L1_MESSAGE_QUEUE_CONTRACT_ADDRESS: Address =
+    address!("0x0d7E906BD9cAFa154b048cFa766Cc1E54E39AF9B");
+
+/// The [`Filter`] used by the [`crate::L1Watcher`] to index events relevant to the rollup node.
+pub static L1_WATCHER_LOG_FILTER: LazyLock<Filter> = LazyLock::new(|| {
```
greged93, line 16: if we add the hook solution in the L1 watcher, this can go

crates/watcher/src/lib.rs
```diff
@@ -0,0 +1,809 @@
...
+    l1_state: L1State,
+    /// The latest indexed block.
+    current_block_number: BlockNumber,
+    /// The sender part of the channel for [`L1Notification`].
+    sender: mpsc::Sender<Arc<L1Notification>>,
+}
+
+/// The L1 notification type yielded by the [`L1Watcher`].
+#[derive(Debug, PartialEq, Eq)]
+pub enum L1Notification {
```
greged93, line 78: This needs to evolve based on the primitives PR discussion

frisitano (commented):
Excellent job on this! I've left a few comments inline. Let me know what you think.

crates/primitives/src/batch.rs
greged93: hmm that's true, we can leave them for now then

crates/watcher/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            return
+        }
+
+        let tail_block = self.unfinalized_blocks.back().expect("tail exists");
+        if tail_block.number < finalized.number {
+            // drain all, the finalized block is past the tail.
+            tracing::trace!(target: "scroll::watcher", tail = ?tail_block.number, finalized = ?finalized.number, "draining all unfinalized blocks");
+            let _ = self.unfinalized_blocks.drain(0..);
+            return
+        }
```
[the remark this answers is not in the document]
greged93: I think this is indeed an edge case, we should actually never have `tail_block.number < finalized.number` if we index without missing any latest blocks.

crates/watcher/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            self.unfinalized_blocks = chain.into();
+        }
+
+        // Update the state and notify on the channel.
+        tracing::trace!(target: "scroll::watcher", number = ?latest.number, hash = ?latest.hash, "new block");
+        self.l1_state.head = latest.number;
+        self.notify(L1Notification::NewBlock(latest.number)).await;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
greged93: so you would move the whole logic of `fetch_unfinalized_chain` here? It might make the function a little heavy to read no? It would be more efficient for sure but remember that the `unfinalized_blocks` should never reach a length longer than 96, so I don't think this will ever be a bottleneck, hence why I prioritized readability of the code over efficiency.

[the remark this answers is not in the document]
frisitano: hmm you're right about making the function heavy. We could possibly modify it to be a function that attaches the provided latest block to the chain returning the block number of the reorg, something like this psuedo code:
```rust
    /// Fetches the chain of unfinalized blocks up to and including the latest block, ensuring no
    /// gaps are present in the chain.
    #[tracing::instrument(target = "scroll::watcher", skip_all)]
    async fn attach_chain_tip(
        &mut self,
        finalized: &Header,
        latest: &Header,
    ) -> L1WatcherResult<Option<u64>> {
        let mut current_block = latest.clone();
        let mut chain = vec![];

        // loop until we find a block contained in the chain, connected to finalized or latest is
        // finalized.
        let (mut chain, pos) = loop {
            if let Ok(pos) =
                self.unfinalized_blocks.binary_search_by(|h| h.number.cmp(&current_block.number))
            {
                debug_assert!(self.unfinalized_blocks[pos].hash == current_block.hash);
                break (chain, Some(pos));
            } else if current_block.parent_hash == finalized.hash ||
                current_block.hash == finalized.hash
            {
                break (chain, None);
            }
            let current_block_number = current_block.number;
            chain.push(current_block);

            tracing::trace!(target: "scroll::watcher", number = ?(current_block_number - 1), "fetching block");
            let block = self
                .execution_provider
                .get_block((current_block_number - 1).into())
                .await?
                .ok_or(EthRequestError::MissingBlock(current_block_number - 1))?;
            current_block = block.header;
        };

        // order new chain from lowest to highest block number.
        chain.reverse();

        // reorg the chain if needed and compute the reorg block number.
        let reorg_block_number = if let Some(pos) = pos {
            // drain all blocks after the fork block.
            tracing::trace!(target: "scroll::watcher", "draining range {:?}", 0..=pos);
            self.unfinalized_blocks.drain(pos + 1..);
            self.unfinalized_blocks.extend(chain);
            Some(self.unfinalized_blocks[pos].number)
        } else {
            self.unfinalized_blocks = chain.into();
            None
        };

        return Ok(reorg_block_number);
    }
```

We would then have less logic in `handle_latest_block` regarding reorg identification. I'm not strongly opinionated on this so feel free to keep as is if you prefer.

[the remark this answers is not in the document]
greged93: I did not take stalled finality into consideration. This would mean we need to add a purge of `self.unfinalized_blocks` any time its length goes over a certain limit, where would you add this?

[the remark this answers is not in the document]
frisitano: Good call at purging `self.unfalized_blocks` I think that would resolve the concern. We could consider wrapping the `VecDeque` data structure into something like:

```rust
pub struct BoundedVecDeque<T> {
    inner: VecDeque<T>,
    capacity: usize,
}
```

Alternatively, we could use something that has constant capacity and supports dropping the oldest value upon `push_back` invocation - [`arraydeque`](https://crates.io/crates/arraydeque) could be a good candidate. 

Implementing `BoundedVecDeque` is probably better as we don't needlessly introduce additional dependencies and is probably simpler. What do you think?

crates/watcher/src/lib.rs
```diff
@@ -0,0 +1,819 @@
...
+        loop {
+            // step the watcher.
+            let _ = self
+                .step()
+                .await
+                .inspect_err(|err| tracing::error!(target: "scroll::watcher", ?err));
+
+            // update loop interval if needed.
+            let loop_interval =
+                if self.is_synced() { SLOW_SYNC_INTERVAL } else { FAST_SYNC_INTERVAL };
```
frisitano, line 151: Do we need to sleep if we are not synced (`self.is_synced()`)? Shouldn't we try to execute the next step as soon as possible? The notification channel is bounded, so it should block if it is at capacity. 

```suggestion
            // sleep if we are already synced
            if self.is_synced() {
                tokio::time::sleep(SLOW_SYNC_INTERVAL).await;
            };
```

crates/watcher/src/lib.rs
```diff
@@ -0,0 +1,819 @@
...
+    /// queries on the RPC provider.
+    execution_provider: EP,
+    /// The buffered unfinalized chain of blocks. Used to detect reorgs of the L1.
+    unfinalized_blocks: VecDeque<Header>,
+    /// The L1 state info relevant to the rollup node.
+    l1_state: L1State,
+    /// The latest indexed block.
+    current_block_number: BlockNumber,
+    /// The sender part of the channel for [`L1Notification`].
+    sender: mpsc::Sender<Arc<L1Notification>>,
```
frisitano, line 73: Should we use a broadcast channel instead for sending events? Take a look at [`reth-tokio-util::EventSender`]( https://github.com/paradigmxyz/reth/blob/main/crates/tokio-util/src/event_sender.rs), I think it would be suitable here.

frisitano (commented):
Added a few more comments inline.

crates/watcher/src/lib.rs
frisitano: Looking into this a bit more the broadcast channel will drop messages if it reaches capacity so it may not be suitable here.

crates/watcher/src/lib.rs
greged93: I wanted to avoid getting rate limited but would be better to remove this and have the rate limit be part of the Provider. Will update.

crates/watcher/src/lib.rs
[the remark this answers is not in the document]
greged93: sounds good, will push an update with it

PR #34, opened 2025-03-13
Initial Database Implementation

frisitano:
# Overview
This PR introduces a `Database` implementation responsible for reading and writing information to the database. We leverage [`seaORM`](https://www.sea-ql.org/SeaORM/) which is an async rust based ORM. It has support for `mysql`, `postgres` and `sqlite`. I would propose that we use an on disk `sqlite` database in the first instance.

# Implementation
We have 4 primary rust objects of interest, these are `Database`, `DatabaseTransaction`, `DatabaseOperations` and `DatabaseConnectionProvider`. The `Database` and `DatabaseTransaction` implement the `DatabaseConnectionProvider` trait which allows for a connection to the database to be sourced. We then have a blanket implementation of `DatabaseOperations`, which defines various CRUD and streaming operations, for all types which implement `DatabaseConnectionProvider`. 

# Migrations
We also introduce a `scroll-migration` crate which is responsible for managing the rollup node migration. 

# Considerations
Moving towards to we should standardize the naming of our crates. We have both `scroll-*` and `rollup-node-*`. We should consider if the two crates introduced in this PR should be `scroll-*` or `rollup-node-*`, I'm leaning towards the latter.


crates/database/db/src/operations.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Get an iterator over all [`BatchInput`]s in the database.
+    async fn get_batch_inputs<'a>(
+        &'a self,
+    ) -> Result<impl Stream<Item = Result<BatchInput, DbErr>> + 'a, DbErr> {
+        Ok(models::batch_input::Entity::find()
+            .stream(self.get_connection())
+            .await?
+            .map(|res| res.map(Into::into)))
+    }
```
greged93, line 82: how does streaming work with Sea ORM? does it provide a snapshot of the DB you can iterate over? meaning any batches added after the call won't be included? Just wondering if we might want to make functions that modify the DB take an exclusive reference to the instance.

greged93 (commented):
just a couple of questions, otherwise LGTM

crates/database/db/src/models/batch_input.rs
```diff
@@ -0,0 +1,105 @@
...
+/// A database model that represents a batch input.
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "batch_input")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    index: i64,
+    version: u8,
+    codec_version: u8,
+    hash: Vec<u8>,
+    block_number: i64,
```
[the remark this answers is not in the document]
frisitano: That would be preferable but sqlite doesn't support u64. `i64::MAX` limit is sufficiently large to store any reasonable block number.

crates/database/db/src/models/batch_input.rs
```diff
@@ -0,0 +1,105 @@
...
+    parent_batch_header: Vec<u8>,
+    #[sea_orm(column_type = "JsonBinary")]
+    chunks: Chunks,
+    skipped_l1_message_bitmap: Vec<u8>,
+    blob_hash: Vec<u8>,
+    finalized_block_number: Option<i64>,
```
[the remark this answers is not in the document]
frisitano: see answer above

crates/database/db/src/models/l1_message.rs
```diff
@@ -0,0 +1,55 @@
...
+use scroll_alloy_consensus::TxL1Message;
+use sea_orm::{entity::prelude::*, ActiveValue};
+
+/// A database model that represents a L1 message.
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "l1_message")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    queue_index: i64,
+    block_number: i64,
```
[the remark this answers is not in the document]
frisitano: see answer above

crates/database/db/src/operations.rs
frisitano: From my understanding, the stream mechanism invokes a `SELECT` statement on the database at a specific point in time. It then iteratively streams the results (rows) one by one. Any changes made after the `SELECT` statement invocation are not included in the result. That's a reasonable suggestion about having an exclusive reference, however, I wonder if this should be handled at the database level using database transactions rather than having a lock at the application level.

greged93: >  I wonder if this should be handled at the database level using database transactions rather than having a lock at the application level.

what type of update does that require to the current code? do you want to open an issue to track this as an improvement for milestone 5?

frisitano: We already have database transaction support implemented in this PR [here](https://github.com/scroll-tech/rollup-node/blob/feat/database-impl/crates/database/db/src/transaction.rs). I will open an issue so that we can track if we also need an application-level `RwLock`. If we wanted to introduce a `RwLock` it would require adding [`parking_lot::RwLock`](https://amanieu.github.io/parking_lot/parking_lot/struct.RwLock.html) to the database such that we have `Arc<RwLock<Database>`.

PR #35, opened 2025-03-14
Indexer Component

frisitano:
# Overview
This PR introduces an `Indexer` component. The indexer is responsible for maintaining state in the database. This state is subsequently used by downstream consumers such as the `Sequencer` and `DerivationPipeline`.


greged93 (commented):
this feels like we are transitioning from the RN manager being the async task executor to the tokio executor. Is this the desired outcome?

greged93 (approved):
lgtm!

PR #38, opened 2025-03-20
feat: codec

### LINKED ISSUE #25: Implement Codec
# Overview
We should implement codecs that allow us to decode batches from L1. There have been several versions of the codec which we must support. See reference [here](https://github.com/scroll-tech/da-codec).

greged93:
Implements the decoding for the Scroll codec, from version 0 to 7.

Resolves #25


frisitano (approved):
This looks great! 

I agree with your suggestion of returning a `BatchInput` instead of a `CommitPayload`, however, I suggest we do that in a follow up PR. We can consider replacing `chunks` with a `CommitPayload`. When implementing that change we should be mindful of the data model. It would be useful if you could add columns to the database model for the `BatchInput`  such as `num_l1_messages`, `l1_message_queue_start_index` so that we can easily map between batches and l1 messages.

PR #41, opened 2025-03-26
feat: integrate batch changes

greged93:
Integrate batch changes in L1Watcher, Database and derivation pipeline.


crates/watcher/src/lib.rs
```diff
@@ -299,43 +297,50 @@ where
...
-            });
+            })
+            .collect::<Vec<_>>();
+
+        // sort the commits and group by tx hash.
+        commit_logs_with_tx
+            .sort_by(|(_, data_a, _), (_, data_b, _)| data_a.batch_index.cmp(&data_b.batch_index));
+        let groups = commit_logs_with_tx.into_iter().chunk_by(|(_, _, hash)| *hash);
+        let groups: Vec<_> =
+            groups.into_iter().map(|(hash, group)| (hash, group.collect::<Vec<_>>())).collect();
```
greged93, line 315: idea here is to order by batch index and then group the commits by transaction hash. This allows us to only query the transaction once when receiving multiple commit events, which can happen after Euclid.

crates/codec/src/lib.rs
```diff
@@ -80,3 +80,21 @@ pub trait CommitDataSource {
     /// Returns the blob for decoding.
     fn blob(&self) -> Option<&Blob>;
 }
+
+/// Returns the codec version from the calldata.
+fn get_codec_version(calldata: &[u8]) -> Option<u8> {
+    const CODEC_VERSION_OFFSET_START: usize = 4;
+    const CODEC_VERSION_OFFSET_END: usize = 4 + 32;
```
frisitano, line 87:
```suggestion
    const CODEC_VERSION_OFFSET_START: usize = 4;
    const CODEC_VERSION_LEN: usize = 32;
    const CODEC_VERSION_OFFSET_END: usize = CODEC_VERSION_OFFSET_START + CODEC_VERSION_LEN;
```

crates/codec/src/lib.rs
```diff
@@ -80,3 +80,21 @@ pub trait CommitDataSource {
...
+    const HIGH_BYTES_FLAG: U256 =
+        U256::from_limbs([u64::MAX, u64::MAX, u64::MAX, 0xffffffffffffff00]);
+    const LOW_BYTES_FLAG: U256 = U256::from_limbs([0, 0, 0, 0xff]);
+
+    let version = calldata.get(CODEC_VERSION_OFFSET_START..CODEC_VERSION_OFFSET_END)?;
+    let version = U256::from_be_slice(version);
+
+    if (version & HIGH_BYTES_FLAG) != U256::ZERO {
+        return None
+    }
```
frisitano, line 97: Why favour `None` over error?

crates/codec/src/lib.rs
```diff
@@ -80,3 +80,21 @@ pub trait CommitDataSource {
...
+
+    Some((version & LOW_BYTES_FLAG).saturating_to())
```
frisitano, line 99: Why use saturating here?  Can we use `to().expect("masked to a single byte")` or something like this?

crates/database/db/src/models/batch_input.rs
```diff
@@ -1,23 +1,18 @@
...
-    codec_version: u8,
     hash: Vec<u8>,
     block_number: i64,
-    parent_batch_header: Vec<u8>,
-    #[sea_orm(column_type = "JsonBinary")]
-    chunks: Chunks,
-    skipped_l1_message_bitmap: Vec<u8>,
-    blob_hash: Vec<u8>,
+    calldata: Vec<u8>,
+    blob_hash: Option<Vec<u8>>,
```
frisitano, line 15: I wonder if we want to keep both encoded and decoded information in the database. i.e. just add `calldata` and `blob_hash` in addition to what we already have. It may be insightful to keep the additional information. Lets open a PR to review the data model in the future.

crates/database/db/src/operations.rs
```diff
@@ -1,27 +1,28 @@
...
 /// The [`DatabaseOperations`] trait provides methods for interacting with the database.
 #[async_trait::async_trait]
 pub trait DatabaseOperations: DatabaseConnectionProvider {
-    /// Insert a [`BatchInput`] into the database.
-    async fn insert_batch_input(&self, batch_input: BatchInput) -> Result<(), DatabaseError> {
-        tracing::trace!(target: "scroll::db", batch_hash = ?batch_input.batch_hash(), batch_index = batch_input.batch_index(), "Inserting batch input into database.");
+    /// Insert a [`BatchCommitData`] into the database.
+    async fn insert_batch(&self, batch_input: BatchCommitData) -> Result<(), DatabaseError> {
+        tracing::trace!(target: "scroll::db", batch_hash = ?batch_input.hash, batch_index = batch_input.index, "Inserting batch input into database.");
         let batch_input: models::batch_input::ActiveModel = batch_input.into();
```
frisitano, line 15: Should we rename `models::batch_input` to `models::batch_commit`?

crates/derivation-pipeline/src/lib.rs
```diff
@@ -4,35 +4,71 @@
...
+    fn set_index_cursor(&mut self, index: u64);
+    /// Set the hash cursor for the provider.
+    fn set_hash_cursor(&mut self, hash: B256);
+}
+
+/// An instance of the trait can be used to provide L1 data.
+pub trait L1Provider: L1MessageProvider {
+    /// Returns corresponding blob data for the provided hash.
+    fn blob(&self, hash: B256) -> Option<Blob>;
 }
```
frisitano, line 42: Is there a reason that `blob` is at the root level and `L1MessageProvider` methods are not? How about:

```suggestion
pub trait L1Provider: L1MessageProvider + L1BlobProvider {}
impl L1Provider for T where T: L1MessageProvider + L1BlobProvider {}
```

frisitano (commented):
Logically looks good. Added a few comments in line.

crates/codec/src/lib.rs
greged93: `saturating_to` doesn't introduce any overhead compared to `to()`, but will update just to stay consistent and clearly translate that no saturation can happen.

crates/database/db/src/models/batch_input.rs
greged93: https://github.com/scroll-tech/rollup-node/issues/44

crates/codec/src/lib.rs
```diff
@@ -77,3 +77,24 @@ pub trait CommitDataSource {
...
+    const LOW_BYTES_FLAG: U256 = U256::from_limbs([0, 0, 0, 0xff]);
+
+    let version = calldata
+        .get(CODEC_VERSION_OFFSET_START..CODEC_VERSION_OFFSET_END)
+        .ok_or(DecodingError::Eof)?;
+    let version = U256::from_be_slice(version);
+
+    if (version & HIGH_BYTES_FLAG) != U256::ZERO {
+        return Err(DecodingError::MissingCodecVersion)
+    }
```
frisitano, line 97: Is `DecodingError::MissingCodecVersion` the correct error here? According to my interpretation of this code, we are checking to ensure that the upper 31 bytes are zero. If the upper 31 bytes are not zero then we have an error. Maybe something like `MalformedCodecVersion` would be more appropriate?

crates/codec/src/lib.rs
```diff
@@ -77,3 +77,24 @@ pub trait CommitDataSource {
...
+
+    Ok((version & LOW_BYTES_FLAG).uint_try_to().expect("masked to a single byte"))
```
frisitano, line 99: In the check above, we have already asserted that the upper 31 bytes are 0. So, is there any value in using the `LOW_BYTES_FLAG`? 

nit: typically would use the terminology of `bitmask` instead of `flag`

crates/codec/src/lib.rs
frisitano: note: bit shifting by 8 bits may be more efficient than using a 256-bit bitmask.

crates/watcher/src/lib.rs
```diff
@@ -299,43 +297,50 @@ where
...
-                        .with_parent_batch_header(info.parent_batch_header())
-                        .with_chunks(info.chunks().map(|c| c.iter().map(|c| c.to_vec()).collect()))
-                        .with_skipped_l1_message_bitmap(info.skipped_l1_message_bitmap())
-                        .with_blob_hashes(blob_hashes);
-
-                // if builder can build a batch input from data, notify via channel.
-                if let Some(batch_input) = batch_builder.try_build() {
-                    self.notify(L1Notification::BatchCommit(batch_input)).await;
-                }
+                let batch_index = decoded_log.batch_index.saturating_to();
```
frisitano, line 333: should we use `.to()` or the more explicit version you used previously `.uint_try_to().expect("Uint conversion error")`

frisitano (approved):
Looks good! Left some minor comments inline.

crates/codec/src/lib.rs
greged93: yeah you're right, using the last mask doesn't make sense, will update

crates/codec/src/lib.rs
greged93: > note: bit shifting by 8 bits may be more efficient than using a 256-bit bitmask.

I checked the implementations for U256, I think the mask is more efficient.
- [&](https://docs.rs/ruint/latest/src/ruint/bits.rs.html#484)
- [>>](https://docs.rs/ruint/latest/src/ruint/bits.rs.html#376)

It seems like the shifting involves a couple operations, whereas the masking is just a `bitand_assign` on all limbs, which I think is highly efficient. What do you think?

crates/watcher/src/lib.rs
greged93: I actually kept the saturating to here, because there isn't a check on the value of the index, which caused a panic in testing where index is generated randomly. I would prefer to either keep the saturating or add a check on the value, what do you think?

greged93: > Is it possible for us to bound the value for the batch_index in testing? I don't think our code logic should be driven by our ability to produce representative test data.

It would be easy and I agree with your statement about avoiding code to be driven by test data. Maybe the `expect` makes more sense since this would be an unrecoverable error, linked to a contract issue on the L1.

PR #42, opened 2025-03-27
feat: L1 providers

greged93:
Introduces the providers crate, centralizing the provider traits and their implementations on the crate. Also provides the `OnlineL1Provider` and the `DatabaseL1MessagesProvider` implementations.


crates/database/db/src/models/batch_input.rs
```diff
@@ -31,6 +32,9 @@ impl From<BatchCommitData> for ActiveModel {
             block_number: ActiveValue::Set(
                 batch_input.block_number.try_into().expect("block number should fit in i64"),
             ),
+            block_timestamp: ActiveValue::Set(
+                batch_input.block_timestamp.try_into().expect("block number should fit in i64"),
+            ),
```
greged93, line 37: timestamp of the L1 block is required in order to fetch the blobs for the block.

crates/database/db/src/models/batch_commit.rs
```diff
@@ -33,6 +34,9 @@ impl From<BatchCommitData> for ActiveModel {
             block_number: ActiveValue::Set(
                 batch_commit.block_number.try_into().expect("block number should fit in i64"),
             ),
+            block_timestamp: ActiveValue::Set(
+                batch_commit.block_timestamp.try_into().expect("block number should fit in i64"),
```
frisitano, line 38:
```suggestion
                batch_commit.block_timestamp.try_into().expect("block timestamp should fit in i64"),
```

crates/providers/src/l1/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+                block_timestamp,
+            ))
+        }
+
+        Ok((block_timestamp - self.genesis_timestamp) / self.slot_interval)
+    }
+}
+
+#[async_trait::async_trait]
+impl<P: L1MessageProvider + Sync> L1BlobProvider for OnlineL1Provider<P> {
```
frisitano, line 88: Do we need the `L1MessageProvider` trait bound on `P` for this impl?

crates/derivation-pipeline/src/lib.rs
```diff
@@ -64,80 +45,105 @@ pub fn derive<P: L1Provider>(
...
-            })
-            .collect::<Vec<_>>();
+        let mut txs = Vec::with_capacity(block.context.num_l1_messages as usize);
+        for _ in 0..block.context.num_l1_messages {
+            let l1_message = l1_provider
+                .next_l1_message()
+                .await
+                .map_err(Into::into)?
+                .ok_or(DerivationPipelineError::MissingL1Message)?;
+            let mut bytes = Vec::new();
```
frisitano, line 64:
```suggestion
            let mut bytes = Vec::with_capacity(l1_message.eip2718_encoded_length());
```

frisitano (approved):
Looks good, left some minor comments inline.

PR #49, opened 2025-04-03
feat: integrate derivation pipeline

greged93:
Integrates the derivation pipeline in the RNM. Adds a `DerivationPipeline` stateful structure, which implements the `Future` trait and yields a `Stream<Item = Result<ScrollPayloadAttributes, DerivationPipelineError>`.


crates/derivation-pipeline/src/lib.rs
```diff
@@ -6,27 +6,161 @@
...
+        // and return.
+        if this.pipeline_futures.is_empty() && this.batch_index_queue.is_empty() {
+            this.waker = Some(cx.waker().clone());
+            return Poll::Pending
+        }
+
+        // if the futures can still grow, handle the next batch.
+        if this.pipeline_futures.len() < MAX_CONCURRENT_DERIVATION_PIPELINE_FUTS {
+            this.handle_next_batch(|queue, fut| queue.push_back(fut));
+        }
```
greged93, line 138: this limitation mechanism could be removed: even we if have up to 1000 pending futures, which all fetch blobs, this would still represent only ~ 128 MB of data.

crates/derivation-pipeline/src/lib.rs
```diff
@@ -6,27 +6,161 @@
...
+                Ok(attributes) => {
+                    this.attributes_queue.extend(attributes);
+                    cx.waker().wake_by_ref();
+                }
+                Err((index, err)) => {
+                    tracing::error!(target: "scroll::node::derivation_pipeline", ?index, ?err, "failed to derive payload attributes for batch");
+                    // retry polling the same batch index.
+                    this.batch_index_queue.push_front(index);
+                    this.handle_next_batch(|queue, fut| queue.push_front(fut));
+                }
```
greged93, line 152: I implemented a sort of retry mechanism here, but maybe this is more suited to be moved to the various providers?

crates/derivation-pipeline/src/lib.rs
frisitano: In the future we should replace the constant with a value from config so we can modify it at runtime.

crates/derivation-pipeline/src/lib.rs
frisitano: Let's leave it as is for now and then replace with provider rety in the future. Lets open an issue to track it for milestone 5.

crates/derivation-pipeline/src/lib.rs
```diff
@@ -6,27 +6,160 @@
...
+                    .get_batch_by_index(index)
+                    .await
+                    .map_err(|err| (index, err.into()))?
+                    .ok_or((index, DerivationPipelineError::UnknownBatch(index)))?;
+
+                derive(batch, provider).await.map_err(|err| (index, err))
+            });
+            queue_fut(&mut self.pipeline_futures, fut);
+        }
+    }
```
frisitano, line 110: nit: I think that this function should return a future instead of executing `queue_fut`. When the caller receives the future they can add it to the back/front of the queue as required. I think `queue_fut` argument makes the code harder to parse. However, not strongly opinionated if you would prefer to keep as is.

crates/node/src/lib.rs
```diff
@@ -73,35 +84,62 @@ pub struct RollupNodeManager<C, EC, P> {
...
         engine: EngineDriver<EC, P>,
+        l1_provider: L1P,
         l1_notification_rx: Option<Receiver<Arc<L1Notification>>>,
         indexer: Indexer,
         forkchoice_state: ForkchoiceState,
         consensus: C,
         new_block_rx: Option<UnboundedReceiver<NewBlockWithPeer>>,
     ) -> Self {
+        let database = indexer.database();
+        let derivation_pipeline = DerivationPipeline::new(l1_provider, database);
```
frisitano, line 131: Should remove the indexer as argument and provide `database: Database` and we can then instantiate the indexer and pipeline here. What do you think?

frisitano (approved):
Looks good, left a few small nits inline. Let me know what you think. CI checks failing.

crates/derivation-pipeline/src/lib.rs
greged93: done https://github.com/scroll-tech/rollup-node/issues/50

crates/watcher/src/lib.rs
```diff
@@ -332,8 +332,19 @@ where
...
+                // slot and query the blobs.
+                let block_timestamp = if log_timestamp.is_none() {
+                    self.execution_provider
+                        .get_block(block_number.into())
+                        .await?
+                        .map(|b| b.header.timestamp)
+                        .ok_or(FilterLogError::MissingBlockTimestamp)?
+                } else {
+                    log_timestamp.expect("checked for Some(...)")
+                };
```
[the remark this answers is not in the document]
frisitano: This looks good to me. I have a slight preference to use `if let` (so we avoid the `expect`) but am not strongly opinionated.

PR #54, opened 2025-04-10
feat: block data

### LINKED ISSUE #52: Extra block data provider
In order to derive the correct block hash, the EN needs to receive the extra information (such as extra data and difficulty). The following action points should be implemented:

- Introduce `BlockDataProvider` which should return the extra information for a block given a block id.
- Populate the extra data from block 0 to 1000 in a file which gets loaded into database during migration.
- Add a `DatabaseExtraDataProvider` which fetches the extra data from database.
- Add a `block_data_hint` field on the `ScrollPayloadAttributes`, which is used to derive the block hash in the `EngineApiTreeHandler`.

greged93:
Adds a migration to the database in order to insert additional block information (extra data and difficulty fields).

Resolves #52.

⚠️ Should be reviewed after https://github.com/scroll-tech/reth/pull/190 is merged.


crates/derivation-pipeline/src/lib.rs
```diff
@@ -155,12 +161,12 @@ where
...
 
-/// Returns a vector of [`ScrollPayloadAttributes`] from the [`BatchCommitData`] and a
-/// [`L1Provider`].
+/// Returns a vector of tuples of [`ScrollPayloadAttributes`] and [`BlockNumber`] from the
+/// [`BatchCommitData`] and a [`L1Provider`].
 pub async fn derive<P: L1Provider>(
     batch: BatchCommitData,
     l1_provider: P,
-) -> Result<Vec<ScrollPayloadAttributes>, DerivationPipelineError> {
+) -> Result<Vec<(ScrollPayloadAttributes, BlockNumber)>, DerivationPipelineError> {
```
greged93, line 169: I need to return the L2 block number in order to get the block data for it

crates/engine/src/engine.rs
```diff
@@ -143,10 +143,10 @@ where
...
             let fc_updated = self.forkchoice_updated(fcs, Some(payload_attributes)).await?;
 
-            // retrieve the execution payload
-            let execution_payload = self
-                .get_payload(fc_updated.payload_id.expect("payload attributes has been set"))
-                .await?;
+            // retrieve the execution payload.
+            let payload_id =
+                fc_updated.payload_id.ok_or(EngineDriverError::MissingExecutionPayloadId)?;
+            let execution_payload = self.get_payload(payload_id).await?;
```
greged93, line 149: This error can actually happen, if you issue a fcs with payload attributes where the unsafe head of fcs is already part of the canonical chain:
https://github.com/scroll-tech/reth/blob/aac5a15729a405197c5dfb53ed09d9b98a763eb6/crates/engine/tree/src/tree/mod.rs#L1178

crates/node/src/lib.rs
```diff
@@ -85,8 +84,8 @@ pub struct RollupNodeManager<C, EC, P, L1P> {
     forkchoice_state: ForkchoiceState,
     /// A collection of pending block imports.
     pending_block_imports: FuturesOrdered<PendingBlockImportFuture>,
-    /// A collection of pending engine driver tasks.
-    pending_engine_tasks: VecDeque<EngineDriverFuture>,
+    /// A pending engine driver task.
+    pending_engine_task: Option<EngineDriverFuture>,
```
greged93, line 88: having a `VecDeque` caused the `EngineDriverFuture` to capture the current fcs of the node manager, hence starting block building on an incorrect head block.

crates/derivation-pipeline/src/lib.rs
frisitano: I suspect this would be needed for the provable pipeline as well?

crates/engine/src/engine.rs
greged93: could you have a case where the `payload_attributes_already_inserted_in_chain` isn't true but the chain is still canonical? Let's say the block import triggers sync of the EN, which syncs to tip, and the L1 consolidation takes longer. At some point, you might issue a FCS with payload attributes that have already been canonicalized by the EN but didn't pass through the engine driver?

frisitano: This would suggest that the execution payload provider is not comprehensive. I would propose that we should ensure the execution payload provider is comprehensive. This can be done by the execution payload provider just being an rpc client of the execution node. In this model `payload_attributes_already_inserted_in_chain` should only be false if we are performing a reorg away from the EN canonical chain. What do you think?

crates/database/migration/src/m20250408_150338_seed_block_data_table.rs
```diff
@@ -0,0 +1,68 @@
+use std::{path::PathBuf, sync::LazyLock};
+
+use alloy_primitives::{Bytes, B256, U256};
+use sea_orm::{prelude::*, ActiveValue, TransactionTrait};
+use sea_orm_migration::{prelude::*, seaql_migrations::Relation};
+
+static BLOCK_DATA_FILE_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
+    let manifest_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    manifest_path.join("./migration-data/block_data.csv")
+});
```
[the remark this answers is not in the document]
greged93: could it bloat the binary size? the current data file is 264KB for 1k blocks, while the release bin size is 62MB. I think it's ok for up to 100k blocks but might get a little much for more, what do you think?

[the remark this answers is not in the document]
frisitano: Yes it certainly will have significant impact on binary size and that will have negative implications for deployment latency as well as other things. However, it will make distribution and integration simpler which I would favour in the first instance. However not strongly opinionated here so if you are then feel free to leave, open an issue and we can asses as we progress.

PR #56, opened 2025-04-14
feat: finalization

### LINKED ISSUE #53: Batch to L2 block
In order to derive the fork choice state of the RN, a mapping from batch to L2 block should be persisted to database. This allows the RN to determine the L2 finalized blocks, based off the finalization status of a batch.

greged93:
Adds a batch to l2 block mapping which is used in the Indexer in order to return the finalized block for the forkchoice state of the rollup node manager.

Resolves #53

### DIFF
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -184,8 +184,8 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
             block_hash = ?block_info.hash,
             "Inserting batch to block into database."
         );
-        let batch_to_block: models::batch_to_block::ActiveModel = (batch_info, block_info).into();
-        models::batch_to_block::Entity::insert(batch_to_block).exec(self.get_connection()).await?;
+        let derived_block: models::derived_block::ActiveModel = (batch_info, block_info).into();
+        derived_block.insert(self.get_connection()).await?;
 
         Ok(())
     }
@@ -205,9 +205,9 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
             .await?;
 
         if let Some(index) = index {
-            Ok(models::batch_to_block::Entity::find()
-                .filter(models::batch_to_block::Column::BatchIndex.lte(index))
-                .order_by_desc(models::batch_to_block::Column::BlockNumber)
+            Ok(models::derived_block::Entity::find()
+                .filter(models::derived_block::Column::BatchIndex.lte(index))
+                .order_by_desc(models::derived_block::Column::BlockNumber)
                 .one(self.get_connection())
                 .await?
                 .map(|model| model.block_info()))
--- a/crates/database/db/src/models/batch_commit.rs
+++ b/crates/database/db/src/models/batch_commit.rs
@@ -20,12 +20,12 @@ pub struct Model {
 /// The relation for the batch input model.
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
 pub enum Relation {
-    /// A one-to-many relation with the batch to block table.
-    #[sea_orm(has_many = "super::batch_to_block::Entity")]
+    /// A one-to-many relation with the derived block table.
+    #[sea_orm(has_many = "super::derived_block::Entity")]
     BatchToBlock,
 }
 
-impl Related<super::batch_to_block::Entity> for Entity {
+impl Related<super::derived_block::Entity> for Entity {
     fn to() -> RelationDef {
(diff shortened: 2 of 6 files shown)


crates/database/db/src/operations.rs
```diff
@@ -146,6 +169,52 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
+        tracing::trace!(
+            target: "scroll::db",
+            batch_hash = ?batch_info.hash,
+            batch_index = batch_info.index,
+            block_number = block_info.number,
+            block_hash = ?block_info.hash,
+            "Inserting batch to block into database."
+        );
+        let batch_to_block: models::batch_to_block::ActiveModel = (batch_info, block_info).into();
+        models::batch_to_block::Entity::insert(batch_to_block).exec(self.get_connection()).await?;
```
greged93, line 188: calling `ActiveModel.insert` can cause a `RecordNotFound` error due to [this issue](https://github.com/SeaQL/sea-orm/issues/2139#issuecomment-2766963383).

crates/indexer/src/lib.rs
```diff
@@ -103,11 +157,40 @@ impl Indexer {
...
+        // finalized the batch.
         database.finalize_batch(batch_hash, block_number).await?;
+
+        // get the finalized block.
+        let finalized_block =
+            Self::fetch_highest_finalized_block(database, batch_hash, l2_block_number).await?;
+
+        // check if the block where the batch was finalized is finalized on L1.
+        let l1_block_number = l1_block_number.lock().await;
+        let finalized_block = finalized_block.filter(|_| *l1_block_number > block_number);
```
frisitano, line 172: should we check if `block_number > l1_block_number` before attempting to fetch the finalized block (`fetch_highest_finalized_block`).

crates/primitives/src/attributes.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use scroll_alloy_rpc_types_engine::ScrollPayloadAttributes;
+
+/// The [`ScrollPayloadAttributes`] coupled with the batch information from which they originated.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ScrollPayloadAttributesWithBatchInfo {
+    /// The payload attributes.
+    pub payload_attributes: ScrollPayloadAttributes,
+    /// The batch information from which the attributes originated.
+    pub batch_info: Arc<BatchInfo>,
```
frisitano, line 12: does this need to be `Arc`?

frisitano (changes requested):
Looks good, I've added some comments inline. I think we may need to modify the `batch_to_block` table as I suspect it currently only has one block per batch.

crates/database/db/src/operations.rs
frisitano: This may be due to the fact that you are using the`batch_index` as primary key and therefore the row is never inserted due to conflicts.

greged93: I think it's not an issue if you use pk auto-increment, which is the case for the other models.

greged93: I will remove the primary key on `batch_index` and check if that works better.

frisitano: I think we should be using `block_number` as the primary key. I think the current implementation has a bug where we will only have one block per batch (the last inserted block most likely).

greged93: still the same issue after updating, I think it's a limitation of the sea orm crate as mentioned in the issue.

crates/database/migration/src/m20250411_072004_add_batch_to_block.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(BatchToBlock::Table)
+                    .if_not_exists()
+                    .col(big_unsigned(BatchToBlock::BlockNumber).primary_key())
```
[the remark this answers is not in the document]
greged93: can we be sure the blocks will always be inserted sequentially? I avoided the pk_auto because of this but I can add it.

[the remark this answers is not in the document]
frisitano: I believe so. Is there a reason to believe they wont?

PR #58, opened 2025-04-15
feat: update initial fcs

### LINKED ISSUE #23: [RNM] Considerations around sync
# Overview
During the initial sync reth uses the `safe` block hash as a sync target as seen in the logic [here](https://github.com/scroll-tech/reth/blob/a337c9777863e3da3ffe9460ab3f0c2fdefc41c7/crates/engine/tree/src/tree/mod.rs#L1227-L1251). This means that we need to source the safe block hash to define the target for the initial sync. An initial proposal is to use the first block received over the wire that passes consensus checks (authorized signer validation) to set the `safe_block_hash`. We currently use the first block to set the `unsafe`, `safe`, and `finalized` hash. Instead, we should leave `finalized` at the genesis hash and wait for the derivation pipeline to provide us with the finalized block hash.
https://github.com/scroll-tech/rollup-node/blob/917eb66676452273bace86b04848ef7e64ea6c8d/crates/node/src/lib.rs#L133-L141

greged93:
Updates the initial FCS of the rollup node to a null hash (0x000...) for all heads. This in turn triggers a backfill up to the current unsafe head at the execution level.

Resolves #23

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,9 +114,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.69"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


frisitano (approved):
LGTM

PR #59, opened 2025-04-16
feat: message queue hash

greged93:
Introduces the message queue hash computation in the indexer. Modifies the L1 watcher in order to fetch the timestamp of L1 messages, allowing for determining if the Euclid fork is active.


crates/primitives/src/transaction.rs
```diff
@@ -1,26 +1,30 @@
...
-pub struct L1MessageWithBlockNumber {
-    /// The L1 block number at which the L1 message was generated.
-    pub block_number: u64,
+pub struct L1MessageEnvelope {
     /// The L1 transaction.
     pub transaction: TxL1Message,
+    /// The L1 block number at which the L1 message was generated.
+    pub block_number: u64,
+    /// The queue hash for the message.
+    pub queue_hash: B256,
```
frisitano, line 12: What do you think about making this an `Option<B256>`? Similarly for the database model making the column nullable.

frisitano (commented):
Looks good. Left a comment inline.

crates/primitives/src/transaction.rs
greged93: that makes a lot of sense! will update

PR #64, opened 2025-04-18
feat: L1 message delay

frisitano:


bin/rollup/src/network.rs
```diff
@@ -153,18 +153,22 @@ where
 
         // Construct the l1 provider.
         let beacon_provider = beacon_provider(l1_provider_args.beacon_rpc_url.to_string());
-        let l1_messages_provider = DatabaseL1MessageProvider::new(db.clone(), 0);
-        let l1_provider =
-            OnlineL1Provider::new(beacon_provider, PROVIDER_BLOB_CACHE_SIZE, l1_messages_provider)
-                .await;
+        let l1_messages_provider = Arc::new(DatabaseL1MessageProvider::new(db.clone(), 0));
```
greged93, line 156: are you sure you want to Arc it and clone it? I think you mean to just clone it, otherwise both the sequencer and the provider will share the same atomic u64 as the message index.

greged93 (commented):
just a remark on the cloning and a question: do you have a design in mind to handle reorgs with the current solution? Otherwise ltgm!

bin/rollup/src/network.rs
frisitano: Great catch, will fix

frisitano:
> do you have a design in mind to handle reorgs with the current solution? Otherwise ltgm!

I'm thinking that we modify the `ReorgIndexed` event to also contain the L1 message index to reorg to `L1ReorgIndexed { block_number: u64, l1_message_queue_index: u64 }`. The indexer will emit this event when it handles a `Reorg` event from the `L1Watcher`. This is an initial idea and we'll need to give it some more thought. I think it'll probably be fine in the first instance however, I'm mindful of your comment about the fact that the `Indexer` has a lot of responsibility at the moment and I tend to agree.

PR #65, opened 2025-04-23
feat: integrate sequencer with node

frisitano:


bin/rollup/src/args.rs
```diff
@@ -23,7 +24,7 @@ pub struct ScrollRollupNodeArgs {
     pub l1_provider_args: L1ProviderArgs,
     /// The sequencer arguments
     #[command(flatten)]
-    pub sequencer_args: Option<SequencerArgs>,
```
greged93, line 26: don't we want to keep this as option? I updated the args in #68 based on https://github.com/clap-rs/clap/issues/5092#issuecomment-1703980717, let me know if you prefer to keep your solution here.

bin/rollup/src/constants.rs
```diff
@@ -9,3 +9,15 @@ pub const PROVIDER_MAX_RETRIES: u32 = 10;
 
 /// The initial backoff for the L1 provider.
 pub const PROVIDER_INITIAL_BACKOFF: u64 = 100;
+
+/// The max L1 messages per block for the sequencer.
```
greged93, line 13: typo in the doc string

bin/rollup/tests/e2e.rs
```diff
@@ -139,13 +139,16 @@ pub async fn build_bridge_node(
         l1_provider_args: L1ProviderArgs {
             l1_rpc_url: None,
             // <https://docs.arbitrum.io/run-arbitrum-node/l1-ethereum-beacon-chain-rpc-providers>
```
greged93, line 141: nit: can also remove the comment here

greged93 (commented):
looks good, thanks for adding a README! left some inline comments

bin/rollup/src/args.rs
frisitano: I'm not sure which solution is preferable at this point. I think your solution is more explicit in that it requires all group values to be provided if one is set, which could prevent an inconsistent config bug (setting block time = 250ms and payload building time defaulting to 500ms).  However, this comes at the expense of simplicity, i.e. user can simply specify `--scroll-sequencer-enabled` and have an operational sequencer with default values. Let's leave this PR as is, then continue the discussion in #68. I will also create an issue to standardise the CLI in milestone 5.

greged93 (approved):
lgtm! CI is failing on clippy

PR #67, opened 2025-04-23
Fix #66: Auto-create database file if it doesn't exist

### LINKED ISSUE #66: [DB] Automatically Create Database File
# Overview
Currently, if the database file has not been created before the node is launched, then the node crashes. Instead we should automatically create the file if it doesn't already exist.

sumitvekariya:
# Auto-create database file if it doesn't exist

## Description
This PR fixes [issue #66](https://github.com/scroll-tech/rollup-node/issues/66) where the node crashes if the database file hasn't been created before launch.

## Changes
- Added directory creation logic to ensure the database parent directory exists
- Added automatic migration execution when connecting to the database
- Added proper error handling and logging for the database initialization process

## Impact
With these changes, the node will automatically create the database file and directory structure if they don't exist, then apply the required migrations to set up the database schema. This improves the user experience by eliminating the need for manual database file creation.

## Testing
- Tested with non-existing database file and directory
- Verified migrations are properly applied to new database
- Ensured existing database files continue to work as before

Fixes #66

### DIFF
--- a/bin/rollup/src/network.rs
+++ b/bin/rollup/src/network.rs
@@ -122,7 +122,7 @@ where
         } else {
             // append the path using strings as using `join(...)` overwrites "sqlite://"
             // if the path is absolute.
-            let path = ctx.config().datadir().db().join("scroll.db");
+            let path = ctx.config().datadir().db().join("scroll.db?mode=rwc");
             "sqlite://".to_string() + &*path.to_string_lossy()
         };
         let db = Database::new(&database_path).await?;
--- a/crates/database/db/src/db.rs
+++ b/crates/database/db/src/db.rs
@@ -2,7 +2,6 @@ use super::{transaction::DatabaseTransaction, DatabaseConnectionProvider};
 use crate::error::DatabaseError;
 
 use sea_orm::{Database as SeaOrmDatabase, DatabaseConnection, TransactionTrait};
-use std::path::Path;
 
 /// The [`Database`] struct is responsible for interacting with the database.
 ///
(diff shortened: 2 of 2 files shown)


frisitano (changes requested):
Thank you very much for your contribution!

Your solution works but I think we should handle this slightly differently following the suggestion here https://github.com/SeaQL/sea-orm/discussions/283#discussioncomment-1564939. I would propose that you revert the current changes and instead we modify [this line](https://github.com/scroll-tech/rollup-node/blob/add7d20bd429cb5aae212c563246e5dce437c4ae/bin/rollup/src/network.rs#L118) to be:
```rust
            let path = ctx.config().datadir().db().join("scroll.db?mode=rwc");
```

This allows for custom database backends to be used via custom `config.database_path` and provides desired behaviour for default case. Would be great if you can test this out please.

sumitvekariya:
> Thank you very much for your contribution!
> 
> Your solution works but I think we should handle this slightly differently following the suggestion here [SeaQL/sea-orm#283 (comment)](https://github.com/SeaQL/sea-orm/discussions/283#discussioncomment-1564939). I would propose that you revert the current changes and instead we modify [this line](https://github.com/scroll-tech/rollup-node/blob/add7d20bd429cb5aae212c563246e5dce437c4ae/bin/rollup/src/network.rs#L118) to be:
> 
> ```rust
>             let path = ctx.config().datadir().db().join("scroll.db?mode=rwc");
> ```
> 
> This allows for custom database backends to be used via custom `config.database_path` and provides desired behaviour for default case. Would be great if you can test this out please.

Hey udnerstood the (rwc) Read, Write & Create mode for sqlite file creation, reverted my changes and updated with your suggestion, Thanks

frisitano (approved):
Thanks for this!

PR #68, opened 2025-04-23
feat: consensus

### LINKED ISSUE #60: [Consensus] Implement Consensus
# Overview
We must implement a consensus implementation that is production-ready. The current Consensus implementation is effectively a no-op.

greged93:
This PR implements changes related to the consensus verifications:

- Add a valid implementation of the PoA consensus.
- Add a dev CLI argument that for now allows to run the node without any consensus checks.
- Fixes an issue with the update of the FCS.

Resolves #60


crates/engine/src/driver.rs
```diff
@@ -65,7 +65,7 @@ where
```
greged93, line 65: I know we talked about reintroducing these lines of code:
```rust
        if self.forkchoice_state.is_genesis() {
            let block_num_hash = block.parent_num_hash();
            self.forkchoice_state = ForkchoiceState::from_block_info(BlockInfo {
                number: block_num_hash.number,
                hash: block_num_hash.hash,
            });
        }
```
and I tried running the node with them, but I didn't see any trigger of the pipeline sync. I think what happened is that it hits this path in the match, meaning it doesn't trigger a sync. My opinion is to keep the code as is, and implement a EL sync and CL sync similar to what optimism are doing.

bin/rollup/src/network.rs
```diff
@@ -90,19 +90,22 @@ where
...
         let engine_api = ScrollAuthEngineApiProvider::new(
             auth_secret,
             self.config.engine_api_url.unwrap_or(format!("http://localhost:{auth_port}").parse()?),
         );
-        let fcs =
-            ForkchoiceState::head_from_genesis(ctx.config().chain.genesis_header().hash_slow());
+        let payload_provider = NoopExecutionPayloadProvider;
+
+        let chain = ctx.config().chain.clone();
+        let fcs = if let Some(named) = chain.chain.named() {
```
frisitano, line 104: `NamedChain` is `Copy` so I don't think we need the clone here.
```suggestion
        let fcs = if let Some(named) = ctx.config().chain.chain.named() {
```

crates/node/src/lib.rs
```diff
@@ -64,7 +63,7 @@ pub struct RollupNodeManager<C, EC, P, L1P, L1MP, CS> {
     /// An indexer used to index data for the rollup node.
     indexer: Indexer<CS>,
     /// The consensus algorithm used by the rollup node.
-    consensus: C,
+    consensus: Box<dyn Consensus + Send>,
```
frisitano, line 66: Should we constrain the `Consensus` trait to be `Send + Debug` instead?

```rust
pub trait Consensus: std::fmt::Debug + Send {
    /// Validates a new block with the given signature.
    fn validate_new_block(
        &self,
        block: &ScrollBlock,
        signature: &Signature,
    ) -> Result<(), ConsensusError>;
}
```
```suggestion
    consensus: Box<dyn Consensus>,
```

crates/node/src/lib.rs
```diff
@@ -85,7 +84,7 @@ impl<C: Debug, EC: Debug, P: Debug, L1P: Debug, L1MP: Debug, CS: Debug> Debug
             .field("derivation_pipeline", &self.derivation_pipeline)
             .field("l1_notification_rx", &self.l1_notification_rx)
             .field("indexer", &self.indexer)
-            .field("consensus", &self.consensus)
+            .field("consensus", &"Consensus")
```
frisitano, line 87: If `Consensus` is `Debug` then we can revert this.

crates/scroll-wire/src/protocol/proto.rs
```diff
@@ -30,7 +32,7 @@ pub struct NewBlock {
 impl NewBlock {
     /// Returns a [`NewBlock`] instance with the provided signature and block.
     pub fn new(signature: Signature, block: reth_scroll_primitives::ScrollBlock) -> Self {
-        Self { signature: Bytes::from(signature.serialize_compact().to_vec()), block }
+        Self { signature: Bytes::from(Into::<Vec<u8>>::into(signature)), block }
```
frisitano, line 35: This is fine for now but in the future I would like to have a solution that ensures correct typing such that we have:
```rust
/// A message that is used to announce a new block to the network.
#[derive(Clone, Debug, PartialEq, Eq, RlpEncodable, RlpDecodable)]
pub struct NewBlock {
    /// The signature from the block author.
    pub signature: Signature,
    /// The block that is being announced.
    pub block: reth_scroll_primitives::ScrollBlock,
}
```

The reason we currently don't have this is because `Signature` does not implement `Encodable`. We may be able to address this by implementing `Encodable` ourselves on this type instead of leveraging the macro but I'm not sure. I will open an issue.

frisitano (approved):
Looks good. Left some minor nits inline. Can you open an issue so that we can update the logic to track signer rotation on L1 system contract please?

frisitano (approved):
Please update README to reflect new CLI args before merge.

crates/scroll-wire/src/protocol/proto.rs
frisitano: #71

bin/rollup/src/args.rs
```diff
@@ -6,6 +6,9 @@ use std::path::PathBuf;
 /// A struct that represents the arguments for the rollup node.
 #[derive(Debug, clap::Args)]
 pub struct ScrollRollupNodeArgs {
+    /// Whether the rollup node should be run in dev mode.
+    #[arg(long)]
+    pub dev: bool,
```
frisitano, line 11: Does this conflict with reth's `dev` CLI arg?

bin/rollup/src/network.rs
greged93: nice catch, didn't notice if was `Copy`.

crates/scroll-wire/src/protocol/proto.rs
greged93: it seems like the implementation of `Encodable` and `Decodable` was removed in this PR. Maybe we can ask for it to be reintroduced?
https://github.com/alloy-rs/core/commit/37e08b23aeeac5db718c2c6bf090c9d5a24f6a8d#diff-79b82cfeeba5106f10ab06fb251c628206b2d8f648bcd94e6fe01c6091763d1c

bin/rollup/src/args.rs
greged93: yes sorry, I changed it to `test` but it reverted for some reason.

crates/scroll-wire/src/protocol/proto.rs
frisitano: Yes that's a good idea. Would you be able to open an issue upstream for this please?

greged93: done!

bin/rollup/src/network.rs
```diff
@@ -139,8 +142,30 @@ where
...
-                .http(l1_rpc_url);
-            let provider = ProviderBuilder::new().on_client(client);
+                .http(url);
+            Some(ProviderBuilder::new().on_client(client))
+        } else {
+            None
+        };
+
+        // Create the consensus.
+        let consensus: Box<dyn Consensus + Send> = if self.config.test {
```
frisitano, line 152: I think `Send` can be removed here.

greged93: oups indeed

PR #80, opened 2025-04-24
chore: improve database interface

frisitano:
# Overview
This PR removes `sea_orm::DbErr` from the `Database` public interface such that in the error case we always return a `DatabaseError`.


greged93 (approved):
lgtm!

Issue #70, opened 2025-04-24
Review and standardise Rollup Node CLI


frisitano:
# Overview
We should review and standardise the rollup-node CLI to have a clean and consistent interface for node operation.

PR #90, opened 2025-04-28
feat: execution payload provider

greged93:
Add an implementation of the execution payload provider.


bin/rollup/Cargo.toml
```diff
@@ -97,6 +97,8 @@ test-utils = [
     "reth-scroll-node/test-utils",
     "reth-scroll-node/test-utils",
     "rollup-node-providers/test-utils",
+    "scroll-engine/test-utils",
+    "scroll-engine/test-utils",
```
frisitano, line 101: why is `scroll-engine` specified twice here? I think this happened to me when using `zepter` but not sure why.

bin/rollup/src/args.rs
```diff
@@ -22,30 +22,62 @@ pub struct ScrollRollupNodeArgs {
...
+    pub max_retries: u32,
+    /// The initial backoff for the provider.
+    #[arg(long = "l1.initial-backoff", id = "l1_initial_backoff", value_name = "L1_INITIAL_BACKOFF", default_value_t = constants::PROVIDER_INITIAL_BACKOFF)]
+    pub initial_backoff: u64,
+}
+
+#[derive(Debug, Default, clap::Args)]
+pub struct BeaconProviderArgs {
+    /// The URL for the Beacon chain.
+    #[arg(long = "beacon.url", id = "beacon_url", value_name = "BEACON_URL")]
```
frisitano, line 58: I like the scoping of the long form using `namespace.*` we should extend that to all custom args for the rollup-node CLI. No need to do it in this PR but we should keep it in mind for #70

bin/rollup/src/network.rs
```diff
@@ -98,7 +102,40 @@ where
...
+                .http(url);
+            Some(ProviderBuilder::new().on_client(client))
+        } else {
+            None
+        };
+
+        // Get a payload provider
+        let payload_provider: Arc<dyn ExecutionPayloadProvider> =
+            if self.config.test || !ctx.config().rpc.http {
+                Arc::new(NoopExecutionPayloadProvider)
```
frisitano, line 125: Can the rollup node function correctly without a payload provider? From my understanding, if the derivation pipeline is enabled, then a functional payload provider must be provided. Maybe we should consider adding some checks / assertions around that?

crates/engine/src/driver.rs
```diff
@@ -46,7 +46,7 @@ where
     /// [`ExecutionPayloadProvider`].
     pub const fn new(
         client: Arc<EC>,
-        execution_payload_provider: Arc<P>,
+        execution_payload_provider: P,
```
frisitano, line 49: Should we change this to be `Option<None>` and raise an error if L1 consolidation is requested (i.e. derivation pipeline is enabled)? What do you think?

crates/providers/Cargo.toml
```diff
@@ -33,11 +35,21 @@ reqwest = { workspace = true, features = ["json"] }
...
 scroll-db = { workspace = true, features = ["test-utils"] }
+scroll-network.workspace = true
 tokio = { workspace = true, features = ["macros"] }
 
 [features]
-test-utils = ["scroll-db/test-utils"]
+test-utils = [
+    "scroll-db/test-utils",
+    "reth-scroll-node/test-utils",
+    "reth-scroll-node/test-utils",
```
frisitano, line 54: duplicate

crates/providers/src/execution_payload.rs
```diff
@@ -1,15 +1,99 @@
...
 /// Implementers of the trait can provide the L2 execution payload for a block id.
 #[async_trait::async_trait]
-pub trait ExecutionPayloadProvider {
-    /// The error returned by the provider.
-    type Error;
-
+#[auto_impl::auto_impl(Arc)]
+pub trait ExecutionPayloadProvider: Sync + Send {
     /// Returns the [`ExecutionPayload`] for the provided [`BlockId`], or [None].
     async fn execution_payload_by_block(
```
frisitano, line 20: nit: 
```suggestion
    async fn execution_payload_for_block(
```

crates/providers/src/test_utils.rs
```diff
@@ -26,3 +28,17 @@ impl BeaconProvider for MockBeaconProvider {
...
+
+#[async_trait::async_trait]
+impl ExecutionPayloadProvider for NoopExecutionPayloadProvider {
+    async fn execution_payload_by_block(
+        &self,
+        _block_id: alloy_eips::BlockId,
+    ) -> Result<Option<ExecutionPayload>, ExecutionPayloadProviderError> {
+        Ok(None)
+    }
+}
```
frisitano, line 44: We may be able to purge this if we change `EngineDriver::execution_payload_provider` to be `Option<P>`

frisitano (commented):
Looks good. Left a few comments inline. Let me know what you think.

crates/engine/src/driver.rs
greged93: I added the logic in `network.rs` but can move it there you think it fits better.

frisitano (approved):
LGTM (pending CI)

PR #99, opened 2025-05-02
feat: sequencer non blocking

frisitano:
# Overview
This PR modifies the `EngineDriver` such that we do not block other fuures during payload building.

### DIFF
--- a/crates/engine/src/driver.rs
+++ b/crates/engine/src/driver.rs
@@ -1,6 +1,7 @@
 use super::{future::EngineDriverFuture, ForkchoiceState};
-use crate::{future::EngineDriverFutureResult, EngineDriverEvent};
+use crate::{future::EngineDriverFutureResult, EngineDriverError, EngineDriverEvent};
 use futures::{ready, task::AtomicWaker, FutureExt, Stream};
+use reth_scroll_primitives::ScrollBlock;
 use rollup_node_primitives::{BlockInfo, ScrollPayloadAttributesWithBatchInfo};
 use rollup_node_providers::ExecutionPayloadProvider;
 use scroll_alloy_provider::ScrollEngineApi;
@@ -34,7 +35,7 @@ pub struct EngineDriver<EC, P = ()> {
     /// The future of the engine driver.
     future: Option<EngineDriverFuture>,
     /// The receiver for the payload building job.
-    payload_building_handle: Option<JoinHandle<EngineDriverFutureResult>>,
+    payload_building_handle: Option<JoinHandle<Result<ScrollBlock, EngineDriverError>>>,
     /// The waker to notify when the engine driver should be polled.
     waker: AtomicWaker,
 }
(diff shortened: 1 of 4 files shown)


crates/engine/src/driver.rs
```diff
@@ -194,25 +197,53 @@ where
...
+        }
+
         // If we have a payload building request from the sequencer, build a new payload.
         if let Some(payload_attributes) = this.sequencer_payload_attributes.take() {
             let fcs = this.fcs.get_alloy_fcs();
             let client = this.client.clone();
             let duration = this.block_building_duration;
 
-            this.future = Some(EngineDriverFuture::payload_building_job(
+            this.payload_building_handle = Some(tokio::spawn(super::future::build_new_payload(
```
greged93, line 241: do we need to spawn it as a tokio task? would creating a separate future `payload_building_future` and storing it separately amount to the same?

crates/engine/src/future/mod.rs
```diff
@@ -263,9 +259,9 @@ where
 }
 
 /// Builds a new payload from the provided fork choice state and payload attributes.
-async fn build_new_payload<EC>(
+pub(crate) async fn build_new_payload<EC>(
```
greged93, line 262: I'm not 100% sure why we need to split `build_new_payload` into two functions? If we have a `payload_building_future` and keep this function as one, doesn't it accomplish the same objective of releasing the driver? Or does this come from the fact that you want a single task that modifies the FCS, so you build the payload and then update it separatly?

greged93 (approved):
more questions than actual comments, lgtm!

crates/engine/src/future/mod.rs
```diff
@@ -39,16 +39,16 @@ type L1ConsolidationFuture = Pin<
...
     Pin<Box<dyn Future<Output = Result<ScrollBlock, EngineDriverError>> + Send>>;
 
 /// An enum that represents the different types of futures that can be executed by the engine
-/// driver. It can be a block import job, an L1 consolidation job, or a payload building job.
+/// driver. It can be a block import job, an L1 consolidation job, or a new payload job.
 pub(crate) enum EngineDriverFuture {
     BlockImport(BlockImportFuture),
     L1Consolidation(L1ConsolidationFuture),
-    PayloadBuildingJob(PayloadBuildingJobFuture),
+    NewPayloadJob(NewPayloadJobFuture),
```
greged93, line 51: nit: this could be `NewPayload`

crates/engine/src/future/mod.rs
frisitano: When we invoke build new payload, we clone the `fcs`. This is used to invoke the payload building job and the future is stored in `payload_building_future`. Whilst the payload is being built we operate other futures which will modify the `safe` and `finalized` elements of the `fcs`. As such, when we want to send the `fcu` to update the `head` we need to make sure we are using the latest `safe` and `finalized` values. Therefore we return back to the `EngineDriver` fetch the current `fcs` and then emit the fcu in the `handle_new_payload` future. Does this make sense and sound reasonable? 

The alternative would be to change `fcs` to `Arc<Mutex<_>>`, but then we have to deal with acquiring locks for every fcu operation.

greged93: that makes sense, thanks for explaining

PR #104, opened 2025-05-08
Feat/add dockerfile

georgehao:


.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Dockerhub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      # build and push to aws ecr
```
Thegaram, line 32: Why not push to DockerHub?

Thegaram: Ah your right, I missed that.

frisitano: why do we want to store in both dockerhub and ecr instead of just dockerhub?

Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM ubuntu:24.04 AS chef
```
frisitano, line 1: Had you considered using the [`rust`](https://hub.docker.com/_/rust) image

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+FROM chef AS builder
+WORKDIR /app
+COPY --from=planner /recipe.json recipe.json
+RUN cargo chef cook --release --recipe-path recipe.json
+RUN --mount=target=. \
+    cargo build --release --target-dir=/app-target
+
+# Release
+
+FROM ubuntu:24.04
```
frisitano, line 30: Is it standard a standard in scroll to use `ubuntu` as the base image or should we use something leaner?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+RUN apt-get update -y && apt-get upgrade -y
+
+# Install basic packages
+RUN apt-get install build-essential curl wget git pkg-config -y
+# Install dev-packages
+RUN apt-get install libclang-dev libssl-dev llvm -y
+
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
frisitano, line 11: We are not pinning the rust version and always pulling rust latest stable. This may be fine but isn't aligned with the concept of "reproducible builds" as the build will be different depending on the latest version of stable rust at the time the image was built.

frisitano (commented):
Added a few commnets inline.

Dockerfile
georgehao: I do a survey for it, rust image uses debian.  however I think we can change to use rust image

.github/workflows/release.yml
georgehao: @sbaizet-ledger I think @sbaizet-ledger want the esk use aws registry, but haven't changed it

Dockerfile
georgehao: current anchored at 1.86.0

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+FROM chef AS builder
+WORKDIR /app
+COPY --from=planner /recipe.json recipe.json
+RUN cargo chef cook --release --recipe-path recipe.json
+RUN --mount=target=. \
+    cargo build --release --target-dir=/app-target
+
+# Release
+
+FROM rust:1.86.0
```
frisitano, line 23: We probably don't want to use `rust` as the base image in the release layer, as it comes packed with the whole rust toolchain, which is massive. Maybe `debian:bullseye-slim` would be a good alternative for the release layer. I'm not sure

Dockerfile
frisitano: If we do chose to go with `debian:bullseye-slim` then I think you will need to add `ca-certs`:
```
RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates && \
    rm -rf /var/lib/apt/lists/*
```

Dockerfile
frisitano: If we do chose to go with `debian:bullseye-slim` then I think you will need to add `ca-certs`:
```
RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates && \
    rm -rf /var/lib/apt/lists/*
```

georgehao: good idea

georgehao: update [e490c60](https://github.com/scroll-tech/rollup-node/pull/104/commits/e490c60c39e2900f8ad58cf2cc9d494cc5345c8b)

frisitano: whoops, sorry but I suggested the old edition of debian. It should actually be `debian:bookworm-slim`, which is the latest edition of debian.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM chef AS builder
+WORKDIR /app
+COPY --from=planner /recipe.json recipe.json
+RUN cargo chef cook --release --recipe-path recipe.json
+RUN --mount=target=. \
+    cargo build --release --target-dir=/app-target
+
+# Release
+
+FROM debian:bullseye-slim
```
frisitano, line 23: lets change to `debian:bookworm-slim`

frisitano (approved):
Lets change the release image to `debian:bookworm-slim` and then this is good to go.

Dockerfile
georgehao: done

PR #107, opened 2025-05-09
feat: euclid

greged93:
Adds Euclid to the Rollup node:
- Use the `ScrollHeaderTransform`.
- Introduces a temp solution of optimistic syncing, allowing for syncing up to the tip. This should be removed once we implement #23.
- Wipes the extra data from the block in `handle_new_block` if we are passed Euclid V2.

### DIFF
--- a/bin/rollup/src/args.rs
+++ b/bin/rollup/src/args.rs
@@ -16,6 +16,10 @@ pub struct ScrollRollupNodeArgs {
     /// A bool that represents if the scroll wire protocol should be enabled.
     #[arg(long, default_value_t = false)]
     pub enable_scroll_wire: bool,
+    /// A bool that represents whether optimistic sync of the EN should be performed.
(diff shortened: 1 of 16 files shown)


bin/rollup/src/args.rs
```diff
@@ -16,6 +16,10 @@ pub struct ScrollRollupNodeArgs {
     /// A bool that represents if the scroll wire protocol should be enabled.
     #[arg(long, default_value_t = false)]
     pub enable_scroll_wire: bool,
+    /// A bool that represents whether optimistic sync of the EN should be performed.
+    /// This is a temp solution and should be removed when implementing issue #23.
+    #[arg(long, default_value_t = false)]
+    pub optimistic_sync: bool,
```
greged93, line 22: this was mostly added just to still be able to perform a sync on the VM and hence test pass the Euclid fork.

Issue #101, opened 2025-05-03
[Node] Migrate the RNM to `NodeAddons`


frisitano:
# Overview
Currently, we instantiate and configure the RNM in the `ScrollRollupNetworkBuilder`:
https://github.com/scroll-tech/rollup-node/blob/bdfee094997465d5266cdb483c400e96cba9854d/bin/rollup/src/network.rs#L60-L236

Whilst this is functional, it is not a particularly clean solution. We should instead configure the RNM as a node addon:
https://github.com/scroll-tech/rollup-node/blob/bdfee094997465d5266cdb483c400e96cba9854d/bin/rollup/src/main.rs#L32

PR #111, opened 2025-05-13
feat: RN as scroll addon

frisitano:
# Overview
This PR refactors the crate strucutre. We introduce `rollup-node-manager` crate, which now holds the RNM imp, we introduce a `rollup-node` crate where we have the node building logic and a node binary. We migrate the instantiation of the RNM to within `ScrollAddons`. We leverage the handles available in the node context to instantiate components of the RNM. We build on this in future PR's to improve testability.

closes: #101


crates/node/src/add_ons/rollup.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        // Get a payload provider
+        let payload_provider = (self.config.test || !ctx.config.rpc.http).then_some({
+            rpc.rpc_server_handles
+                .rpc
+                .new_http_provider_for()
+                .map(Arc::new)
+                .map(AlloyExecutionPayloadProvider::new)
+                .expect("failed to create payload provider")
+        });
```
greged93, line 108: I think the logic is incorrect here, should be `(!(self.config.test || !ctx.config.rpc.http)).then_some(...)` or `!self.config.test & ctx.config.rpc.http`

crates/node/src/args.rs
```diff
@@ -38,9 +31,25 @@ pub struct ScrollRollupNodeArgs {
...
+/// The network arguments.
+#[derive(Debug, Default, Clone, clap::Args)]
+pub struct NetworkArgs {
+    /// A bool to represent if new blocks should be bridged from the eth wire protocol to the
+    /// scroll wire protocol.
+    #[arg(long = "eth-scroll-bridge.enabled", default_value_t = false)]
+    pub enable_eth_scroll_wire_bridge: bool,
+    /// A bool that represents if the scroll wire protocol should be enabled.
+    #[arg(long = "scroll-wire.enabled", default_value_t = false)]
+    pub enable_scroll_wire: bool,
```
greged93, line 48: I'm wondering if these shouldn't be true on default? I've already lost some time forgetting to set them.

crates/node/tests/e2e.rs
```diff
@@ -123,43 +124,28 @@ pub async fn build_bridge_node(
...
     let node_config = NodeConfig::new(chain_spec.clone())
         .with_network(network_config.clone())
-        .with_rpc({
-            let mut args =
-                RpcServerArgs::default().with_http().with_http_api(RpcModuleSelection::All);
-            args.auth_jwtsecret =
-                Some(PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("assets/jwt.hex"));
-            args
-        })
+        .with_rpc(RpcServerArgs::default().with_http().with_http_api(RpcModuleSelection::All))
```
greged93, line 127: we don't need auth anymore?

greged93 (commented):
Couple of nits, a typo but looks good! Let's get some testing in after we merge this!

crates/manager/src/manager/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+            .and_then(|i| Signature::from_raw(&extra_data[i..]).ok())
+        {
+            trace!(target: "scroll::bridge::import", peer_id = %peer_id, block = ?block.hash_slow(), "Received new block from eth-wire protocol");
+            NewBlockWithPeer { peer_id, block, signature }
+        } else {
+            warn!(target: "scroll::bridge::import", peer_id = %peer_id, "Failed to extract signature from block extra data");
+            return;
+        };
+
+        self.handle_new_block(block);
```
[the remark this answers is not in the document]
frisitano: Currently, this isn't practical because we also invoke the `handle_new_block` method from within the `handle_network_manager_event` method, which receives new blocks from the `ScrollNetworkManager`.

crates/node/src/add_ons/rollup.rs
frisitano: good catch

crates/node/tests/e2e.rs
frisitano: Now we have access to the `rpc_handle` when instantiating the RNM we can delegate the auth configuration to `new_http_provider_for` which doe this for us. Previoulsy this was not possible as we were instantiating the RNM in the `NetworkBuidler` and at that point the `RpcHandle` was not available as it has not been instantiated yet (its instantiated as part of the addons).
https://github.com/scroll-tech/rollup-node/blob/38e9b4836323ff5eb723872f6c8f8f9a851cff34/crates/node/src/add_ons/rollup.rs#L101-L108

greged93 (approved):
lgtm!

PR #113, opened 2025-05-14
feat: consensus signer

### LINKED ISSUE #72: [Consensus] Track system contract signer
Currently, the system contract signer for the `PoAConsensus` is only initialized at start up and as such doesn't track any changes in the signer. Implement tracking of the authorized signer on the L1.

greged93:
Adds an polling update of the consensus configuration on the watcher, propagating updates across the node to reach the consensus.

Resolves #72


crates/node/src/consensus.rs
```diff
@@ -45,33 +36,23 @@ impl Consensus for NoopConsensus {
...
 #[derive(Debug)]
 pub struct PoAConsensus {
-    authorized_signers: Vec<Address>,
+    authorized_signers: HashSet<Address>,
 }
 
 impl PoAConsensus {
     /// Creates a new [`PoAConsensus`] consensus instance with the given authorized signers.
-    pub const fn new(authorized_signers: Vec<Address>) -> Self {
+    pub const fn new(authorized_signers: HashSet<Address>) -> Self {
```
frisitano, line 44: nit: As a convenience, you can make this generic over the collection type. Something like below:
```rust
    pub fn new<T: IntoIterator<Item = Address>>(authorized_signers: T) -> Self {
        Self { authorized_signers: authorized_signers.into_iter().collect() }
    }
```

frisitano (dismissed):
Looks good. I have some questions:

- How does this behave in the context of reorgs?
- How do we remove authorized signers?

greged93:
> How does this behave in the context of reorgs?

For now, the authorized signer will be pulled from the L1 every latest block, meaning reorgs will just cause it to repull the latest signer at the next block. We should consider reorgs when we use logs to fetch signer updates.

> How do we remove authorized signers?

Currently the node just reads the signer from the L1, so removing a signer would require an update of the L1 contract.

frisitano (approved):
LGTM

frisitano:
@greged93 can you open an issue on the smart contracts repo to introduce events for removal and addition of signers please?

greged93:
@frisitano done https://github.com/scroll-tech/scroll-contracts/issues/109

PR #116, opened 2025-05-16
feat: manager handle

frisitano:
Introduces `RollupManagerHandle` and `ScrollAddOnsHandle` and integrates them with the rollup node.


crates/manager/src/manager/mod.rs
```diff
@@ -28,17 +28,26 @@ use std::{
     sync::Arc,
```
greged93, line 28: If I understand correctly, this command pattern should only be used for testing right? Should we place the methods and field behind a `test-utils` flag?

greged93 (commented):
just a small comment, otherwise looks good!

crates/manager/src/manager/mod.rs
frisitano: https://github.com/scroll-tech/rollup-node/issues/119

Issue #93, opened 2025-04-29
[Node] Introduce RNM Builder


frisitano:
# Overview
We should introduce a RNM builder type. This will allow us to assert that the RNM is correctly instantiated.

PR #117, opened 2025-05-16
feat: l1 watcher tx

frisitano:
# Overview
Introduces a channel to publish `L1Notification` to the `RollupManager`. This can be used for integration testing. We include the `Sender` in `ScrollAddonsHandle` when `test-utils` is enabled.


crates/node/src/add_ons/rollup.rs
```diff
@@ -134,12 +135,16 @@ impl RollupManagerAddOn {
             Box::new(poa)
         };
```
greged93, line 136: feeling like this function is getting a little messy, we should probably split it a little, what do you think? Not in this PR but just a general remark.

crates/node/src/add_ons/rollup.rs
```diff
@@ -134,12 +135,16 @@ impl RollupManagerAddOn {
...
-        };
+        let (l1_notification_tx, l1_notification_rx) =
+            if let Some(provider) = provider.filter(|_| !self.config.test) {
+                // Spawn the L1Watcher
+                (None, Some(L1Watcher::spawn(provider, node_config).await))
+            } else {
+                // Create a channel for L1 notifications that we can use to inject L1 messages for
+                // testing
+                let (tx, rx) = tokio::sync::mpsc::channel(1000);
+                (Some(tx), Some(rx))
```
greged93, line 146: is this useful since the `l1_watcher_tx` is behind a feature flag?

greged93 (commented):
looks good! Just a general remark on the `launch` method of the `RollupManagerAddOn` and a comment

crates/node/src/add_ons/rollup.rs
frisitano: Agreed. I think that this will be resolved with the introduction of the RNM builder #93

greged93 (approved):
lgtm!

Issue #127, opened 2025-05-22
[E2E Testing] Introduce `RollupManagerTestContext` trait


frisitano:
# Overview
We have re-architected the rollup node manager to be a node addon. As such, we can now access the rollup node manager handle from the node handle. We use this to invoke methods on the rollup node manager. Currently, we are fetching the rollup manager handle explicitly and invoking commands on the handle. It would be preferable if we had a standard suite of methods that we can invoke that abstract the implementation details. Examples of the explicit / manual mechanism seen below:
https://github.com/scroll-tech/rollup-node/blob/b25c8de23048e5f6a20034cc1ca5c935b62abd48/crates/node/tests/e2e.rs#L31-L158

# Node Test Design 
There are two approaches we can take for the node here:
- Introduce a `RollupManagerTest` trait and implement it on  [`NodeTestContext`](https://github.com/scroll-tech/reth/blob/03b78f4c549d1b593b1246f5d021a854da108219/crates/e2e-test-utils/src/node.rs#L35).
- Wrap `NodeTestContext` in `ScrollNodeTestContext` which includes a `RollupManagerTestContext` field which own a rollup node manager handle and implements useful methods for testing. For convenience, we can implement `deref` on `ScrollNodeTestContext` to get access to methods from `NodeTestContext` for free.

I'm slightly in favor of approach 2.

# Node Network Design
The [`setup_engine`](https://github.com/scroll-tech/rollup-node/blob/b25c8de23048e5f6a20034cc1ca5c935b62abd48/crates/node/src/test_utils.rs#L24) function returns a `Vec<NodeTestContext>`. We can consider introducing a wrapper around this `ScrollNetworkTestContext`, which includes helper methods to invoke actions on nodes in the network and also to assert expected outcomes on the rest of the nodes.

PR #130, opened 2025-05-23
feat: delayed consolidation

### LINKED ISSUE #23: [RNM] Considerations around sync

greged93:
Introduces a delayed consolidation feature: the rollup node will not consume L1 notifications from the watcher as long as the EN isn't synced. This will implicitly pause the watcher once its channel for the notifications will be full.

Resolves #23


crates/engine/src/future/mod.rs
```diff
@@ -292,9 +296,13 @@ where
...
-    Ok(get_payload(client.clone(), fc_updated.payload_id.expect("payload attributes has been set"))
-        .await?
-        .try_into_block()?)
+    let payload = get_payload(
+        client.clone(),
+        fc_updated.payload_id.expect("payload attributes has been set"),
+    )
+    .await?;
+    let data = ExecutionData::new(payload, Default::default());
+    Ok(try_into_block(data, chain_spec)?)
```
greged93, line 305: `try_into_block` will always set the difficulty of the block to 1.

crates/node/src/test_utils.rs
```diff
@@ -86,9 +93,7 @@ where
             })
             .await?;
 
-        let mut node =
-            NodeTestContext::new(node, |_| panic!("should not build payloads using this method"))
-                .await?;
+        let mut node = NodeTestContext::new(node, scroll_payload_attributes).await?;
```
greged93, line 96: I actually need this in order to be able to call `advance` on the test node.

crates/manager/src/manager/mod.rs
```diff
@@ -363,40 +377,61 @@ where
...
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.get_mut();
+        // Helper macro, proceeds with the $task if $proceed is true.
+        macro_rules! proceed_if {
+            ($proceed: expr, $task: expr) => {
+                if $proceed {
+                    $task
+                }
+            };
+        }
```
frisitano, line 387: I'm not sure what is cleaner but instead of this you could add a filter on the object you are trying to operate on. 
Example
```rust
if let Some(sequencer) = this.sequencer.filter(|_| !this.engine.is_syncing()).as_mut() {
    sequencer.build_payload_attributes();
}
```

What do you think?

crates/node/src/test_utils.rs
frisitano: Is this so that you can advance the EN without advancing the rN?

frisitano (commented):
I've provided a preliminary review. I'm unsure about introducing the `Provider` in the `EngineDriver`, I think we should be able to use the `fcs` to get the head block number. Let's resolve this discussion first and then I will review again.

crates/engine/src/driver.rs
```diff
@@ -49,32 +62,31 @@ pub struct EngineDriver<EC, CS, P = ()> {
...
     ) -> Self {
         Self {
             client,
             chain_spec,
-            execution_payload_provider,
+            provider,
             fcs,
             block_building_duration,
-            optimistic_sync,
+            syncing: false,
```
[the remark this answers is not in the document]
greged93: hmmm true, at least we want to avoid issues with the sequencer.

crates/node/src/test_utils.rs
greged93: yes, I could also advance the EN using the RN, setting it to sequencer, but it's easier to just advance it using the implemented `advance` method.

greged93: ahhhh I see what you mean. It's not really desired but it doesn't have any impact on the test, as the node is just used to sync via P2P in this case. Would you rather I advance it using the RN?

frisitano: I'm not strongly opinionated in this case. I think it's fine as is. However, in the future, I would suggest we wrap the `NodeTestContext` and implement convenience methods like `advance` ourselves as described in https://github.com/scroll-tech/rollup-node/issues/127.

crates/node/src/add_ons/rollup.rs
```diff
@@ -104,7 +102,7 @@ impl RollupManagerAddOn {
             ctx.config.chain.clone(),
             payload_provider,
             fcs,
-            true,
+            self.config.engine_driver_args.en_sync_trigger,
```
frisitano, line 105: Above on line 99 I was proposing that we make a request to the EN to get the FCS. That way if the node is shut down and later restarted, the fcs will be in sync with the EN.

frisitano (commented):
Looks good. Left a few comments inline.

crates/node/tests/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+        if retries > 20 || num > node_config.engine_driver_args.en_sync_trigger {
+            break
+        }
+        num = provider.get_block_number().await.unwrap();
+        tokio::time::sleep(tokio::time::Duration::from_millis(200)).await;
+        retries += 1;
+    }
+
+    // Build and announce a new block to have engine driver exit syncing mode.
+    build_block_and_announce(&synced, sequencer_node_config.sequencer_args.block_time).await;
```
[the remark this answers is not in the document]
greged93: This simplified the test a lot thanks!

crates/node/tests/sync.rs
```diff
@@ -37,21 +28,17 @@ async fn test_should_trigger_pipeline_sync_for_execution_node() {
...
-    // Advance the chain.
-    build_blocks(
-        &synced,
-        node_config.engine_driver_args.en_sync_trigger as usize + 1,
-        sequencer_node_config.sequencer_args.block_time,
-    )
-    .await;
+    // Wait for the chain to be advanced by the sequencer.
+    let en_sync_trigger = node_config.engine_driver_args.en_sync_trigger + 1;
+    tokio::time::sleep(tokio::time::Duration::from_millis(en_sync_trigger * block_time)).await;
```
frisitano, line 33: It may be more deterministic to use the event listener to wait for `BlockSequenced` event as seen in the test below
https://github.com/scroll-tech/rollup-node/blob/db1faaa674bfda3f2621942f0398e6e5d9481f1e/crates/node/tests/e2e.rs#L83

frisitano (approved):
Looks good. We could optionally use events instead of time to make our test cases more declarative.

crates/node/tests/sync.rs
```diff
@@ -37,21 +28,17 @@ async fn test_should_trigger_pipeline_sync_for_execution_node() {
...
 
     // Connect the nodes together.
     synced.network.add_peer(unsynced.network.record()).await;
     unsynced.network.next_session_established().await;
     synced.network.next_session_established().await;
 
-    // Announce the latest block.
-    announce_latest_block(&synced).await;
+    // Wait for the unsynced node to receive a block.
+    tokio::time::sleep(tokio::time::Duration::from_millis(block_time)).await;
```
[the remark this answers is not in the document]
greged93: ahhh that's right, was not aware of this, will udpate.

frisitano (approved):
Looks great! LGTM

PR #139, opened 2025-05-27
feat: watcher logs

### LINKED ISSUE #136: [Watcher] Improve query range
- [x] Verify that the logs query range will not miss logs in case of reorgs.
- [x] Verify that the logs query range will not reindex data when we reach the tip of the chain.
- [x] Lower query range (or make adaptative) to avoid query error on logs.

greged93:
Reduce logs block range and avoid reindexing latest block on the L1 watcher.

Resolves #136

### DIFF
--- a/crates/watcher/src/lib.rs
+++ b/crates/watcher/src/lib.rs
@@ -843,8 +843,7 @@ mod tests {
     async fn test_should_handle_l1_messages() -> eyre::Result<()> {
         // Given
         let (finalized, latest, chain) = chain(10);
-        let (watcher, mut receiver) =
-            l1_watcher(chain, vec![], vec![], finalized.clone(), latest.clone());
+        let (watcher, _) = l1_watcher(chain, vec![], vec![], finalized.clone(), latest.clone());
 
         // build test logs.
         let mut logs = (0..10).map(|_| random!(Log)).collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


crates/watcher/src/lib.rs
```diff
@@ -183,18 +183,20 @@ where
...
+            self.handle_batch_commits(&logs).await?;
+            self.handle_batch_finalization(&logs).await?;
+            // TODO(greg): update with logs once system contract emits logs.
+            self.handle_system_contract_update(&latest).await?;
 
-        // update the latest block the l1 watcher has indexed.
-        self.update_current_block(&latest);
+            // update the latest block the l1 watcher has indexed.
+            self.update_current_block(&latest);
+        }
```
frisitano, line 199: Should we make notifications for a range of blocks atomic? Let's consider the following case. We fetch logs for the next range using `let logs = self.next_filtered_logs(latest.header.number).await?`. We then invoke `self.handle_l1_messages(&logs).await?` which works as expected and sends notifications. We then invoke `self.handle_batch_commits(&logs).await?` and this errors. On the `step` invocation, we would then try to process the range again, and this would result in duplicate notifications associated with `self.handle_l1_messages(&logs).await?`. Instead, we can have `self.handle_l1_messages(&logs).await?`, and similar methods, return a vector of notifications, e.g.
```rust
let mut notifications = Vec::new();
notifications.extend(self.handle_l1_messages(&logs).await?);
notifications.extend(self.handle_batch_commits(&logs).await?);
...
for notification in notifications {
    self.notify(notification)
}
```
This way, all notifications for a block range are atomic and would ensure that no duplicated notifications are sent downstream. What do you think?

frisitano (commented):
Looks good, left a comment inline. Let me know what you think.

PR #140, opened 2025-05-27
Sequencer stalling

frisitano:
# Context
When @georgehao was performing benchmarks, he ran into an issue in which the sequencer would stall. After doing some investigation, I suspect this is due to the fact that the `Interval` uses `Burst` for `MissedTickBehaviour`. This results in erroneous block building frequency and gets the manager stuck. 

# Solution
In this PR we modify the `Interval` to use `tokio::time::MissedTickBehavior::Delay`. This should resolve the issue but we should test this in practice. 

Context: https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variants

### DIFF
--- a/crates/manager/src/manager/mod.rs
+++ b/crates/manager/src/manager/mod.rs
@@ -170,8 +170,7 @@ where
             event_sender: None,
             sequencer,
             signer,
-            block_building_trigger: block_time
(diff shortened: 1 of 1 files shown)


greged93 (approved):
lgtm, was not aware of this behavior of the ticker, good to know!

PR #150, opened 2025-06-02
feat: graceful shutdown

frisitano:
# Overview
This PR implements a graceful shutdown for the rollup node to ensure that critical in-flight tasks are driven to completion before shutdown.

# Test strategy
The test implemented for this feature is to start up a rollup node externally to the EN, such that we can manually poll it. This is required such that we can simulate a crash/shutdown midway through the processing of a batch. We simulate a crash midway through processing and then ensure that the node can start up and reprocess the latest batch that we crashed at. We assert that the EN safe block progresses to the expected chain tip upon restart.

Note: There may be an edge case in which the finalized head is greater than the safe head on restart, as we revert to the safe head. As such, when we issue the forkchoice rule, we ensure that we only issue an update if the safe block hash is greater than the finalized block hash. 

# Implementation
The following modifications have been made to support this feature:
- At startup, always reindex the latest block
- Modify database models to do nothing on conflict when inserting new data from the watcher
- Add a method to the database to determine which safe block the node should start at
- Add a utility function to determine the l1 block number the watcher should start at
- Introduce a `ConsolidationOutcome` type to clean up the consolidation interface
- Only update safe head if the safe head is greater than the finalized head
- Add a `run_until_graceful_shutdown` function on the RNM and integrate with the reth `TaskExecutor`
- Migrate the instantiation of the RNM from the `ScrollRollupConfig` 

# Update
On startup, we purge all data that is above the latest persisted finalized L1 block number (we assume that L1 has reorged). We then start up the L1 watcher such that we replay from the batch associated with the latest L2 safe block post-purge.


crates/manager/src/manager/mod.rs
```diff
@@ -511,3 +532,18 @@ fn delayed_interval(interval: u64) -> Interval {
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
     interval
 }
+
+/// Computes the starting block for the watcher from the database.
+pub async fn compute_watcher_start_block_from_database(
```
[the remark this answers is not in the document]
frisitano: No I'll migrate this to `DatabaseOperations`.

crates/database/db/src/operations.rs
```diff
@@ -19,8 +19,16 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
+        match models::batch_commit::Entity::insert(batch_commit)
+            .on_conflict(
+                OnConflict::column(models::batch_commit::Column::Index).do_nothing().to_owned(),
+            )
+            .exec(self.get_connection())
+            .await
+        {
+            Err(sea_orm::DbErr::RecordNotInserted) => Ok(()),
+            x => Ok(x.map(|_| ())?),
+        }
```
frisitano, line 31: This can only ever happen during startup and we already know that the batch is finalized due to our startup `unwind` logic so `do_nothing()` should be valid here.

crates/database/db/src/operations.rs
```diff
@@ -113,7 +152,23 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
+                        models::l1_message::Column::GasLimit,
+                        models::l1_message::Column::To,
+                        models::l1_message::Column::Value,
+                        models::l1_message::Column::Sender,
+                        models::l1_message::Column::Input,
+                    ])
+                    .to_owned(),
+            )
+            .exec(self.get_connection())
+            .await?;
```
frisitano, line 171: Given that we will only ever replay finalized messages, we can possibly change this to `do_nothing()`.

greged93: > Given that we will only ever replay finalized messages,

Can you explain this part? It can happen that we reindex messages that weren't finalized right?

crates/manager/src/manager/mod.rs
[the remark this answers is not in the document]
frisitano: This logic has been removed and is now part of https://github.com/scroll-tech/rollup-node/blob/b637bb0f29a0469f00cd5dcaabd77457702e7ceb/crates/database/db/src/operations.rs#L303

crates/database/db/src/operations.rs
frisitano: I don't think it's possible to have conflicts for unfinalized L1 messages. If we have an L1 reorg, then we first unwind to the reorged L1 block height and remove all L1 messages after this height. Then, when we progress the L1 chain and have new L1 messages for the same L1 message indexes, there will be no entries in the table, as they have been removed. There is only one case (at least given a sound implementation) where we can have conflicts, which is when we replay some finalized L1 blocks on startup.

greged93 (commented):
just a nit and 2 questions, otherwise looks good!

crates/node/src/args.rs
```diff
@@ -46,6 +78,235 @@ impl ScrollRollupNodeConfig {
...
+            .await
+            .unwrap_or_else(chain_spec_fcs);
+
+        let chain_spec = Arc::new(chain_spec.clone());
+
+        // On startup we replay the latest batch of blocks from the database as such we set the safe
+        // block hash to the latest block hash associated with the previous consolidated
+        // batch in the database.
+        let finalized_block_number = db.get_finalized_l1_block_number().await?.unwrap_or(0);
+        rollup_node_indexer::unwind(db.clone(), chain_spec.clone(), finalized_block_number).await?;
```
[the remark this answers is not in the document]
greged93: ohhh ok got it, yeah that makes sense

crates/node/tests/e2e.rs
```diff
@@ -243,3 +250,224 @@ async fn can_bridge_blocks() {
...
+
+    // Lets fetch the first consolidated block event - this should be the first block of the batch.
+    let l2_block = loop {
+        let event = loop_until_event(&mut rnm, &mut rnm_events).await;
+        if let RollupManagerEvent::L1DerivedBlockConsolidated(consolidation_outcome) = event {
+            break consolidation_outcome.block_info().clone();
+        }
+    };
+
+    // Assert that the consolidated block is the first block of the batch.
```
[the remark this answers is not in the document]
frisitano: I'll explain the reasoning for this and then explain why it's challenging to do a good test for this at the moment.

When we start up the node, we remove all unfinalized L1 data. We are then left with safe L2 blocks, which are guaranteed to be associated with batches that are produced in a finalized L1 block. However, we have no guarantee that we have processed all blocks associated with the batch that contains the current latest L2 block (it's possible the node shut down mid batch processing). As such, we want to replay the whole batch starting from the first block of the batch.

In the current test case, we only have 2 batches. We finalize the L1 block associated with the first batch; however, we have no way to guarantee that we have processed all blocks from the first batch, so we start by reconsolidating the whole batch. A better test would be to have three batches. We process and finalize the first two batches and then stop the node midway through the third batch (which isn't L1 finalized). We would then expect that upon startup, we would start by reconsolidating the second batch. It's difficult to implement this test because the third mainnet batch contains L1 messages, which makes setting up the test case a bit awkward, as we would have to include L1 messages. I think it would be quite a lot of effort to do this at the current point in time. How about I open an issue to improve this test case in the future.

I encourage you to validate the logic I mentioned above to ensure it's reasonable, as I may be missing an edge case; however, I've tried to be as "defensive" as possible.

crates/database/db/src/operations.rs
greged93: > An important detail to note here is that on startup, we purge all L1 data that was not finalized based on the latest finalized L1 block number at the time the node last shut down.

Got it, thank you for the explanation.

crates/node/tests/e2e.rs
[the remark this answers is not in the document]
frisitano: https://github.com/scroll-tech/rollup-node/issues/174

crates/database/db/src/operations.rs
```diff
@@ -300,8 +304,13 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
     /// the previous batch (i.e., the batch before the latest safe block). It then returns a tuple
     /// of this latest fetched block and the L1 block number of the batch.
-    async fn get_startup_data(&self) -> Result<(Option<BlockInfo>, Option<u64>), DatabaseError> {
+    async fn get_startup_data(
+        &self,
+        genesis_hash: B256,
+    ) -> Result<(Option<BlockInfo>, Option<u64>), DatabaseError> {
         tracing::trace!(target: "scroll::db", "Fetching startup safe block from database.");
+        let finalized_block_number = self.get_finalized_l1_block_number().await?.unwrap_or(0);
+        self.unwind(genesis_hash, finalized_block_number).await?;
```
greged93, line 313: After thinking about it I wonder if the unwind should happen here, given the name of the method. Maybe we can rename the method? If not I would rather have unwind out of it to avoid unwillingly unwinding the DB.

crates/database/db/src/operations.rs
```diff
@@ -300,8 +304,13 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
     /// L1 block number at which this batch was produced. It then retrieves the latest block for
     /// the previous batch (i.e., the batch before the latest safe block). It then returns a tuple
     /// of this latest fetched block and the L1 block number of the batch.
-    async fn get_startup_data(&self) -> Result<(Option<BlockInfo>, Option<u64>), DatabaseError> {
+    async fn get_startup_data(
+        &self,
+        genesis_hash: B256,
+    ) -> Result<(Option<BlockInfo>, Option<u64>), DatabaseError> {
         tracing::trace!(target: "scroll::db", "Fetching startup safe block from database.");
+        let finalized_block_number = self.get_finalized_l1_block_number().await?.unwrap_or(0);
```
greged93, line 312: should we move this into `unwind`?

crates/database/db/src/operations.rs
frisitano: Agreed, let's rename the method.

frisitano: I've renamed to `prepare_on_startup` such that it can be invoked with `db.prepare_on_startup(..)` but please feel free to suggest alternatives.

greged93 (approved):
lgtm!

Issue #143, opened 2025-05-28
[Sequencer] Introduce finalized `L1MessageProvider` predicate


frisitano:
# Overview
Currently, the predicate that is used to specify the `L1MessageProvider` is based on block depth. We should introduce an additional predicate that only serves finalised L1 messages. This mode should then be configurable via the CLI arguments, such that the user can choose either block depth or finalized mode.

https://github.com/scroll-tech/rollup-node/blob/db1faaa674bfda3f2621942f0398e6e5d9481f1e/crates/sequencer/src/lib.rs#L157-L159

We should update the sequencer CLI arguments such that the user can specify either a block depth at which the transactions should be included, or alternatively, only to include finalized transactions. 

https://github.com/scroll-tech/rollup-node/blob/3e02fc3807ee71548f962b74dfb23a159c07cef3/crates/node/src/args.rs#L95-L113

We then use this configuration when instantiating the sequencer to configure the L1 message inclusion mechanism we should use:

https://github.com/scroll-tech/rollup-node/blob/3e02fc3807ee71548f962b74dfb23a159c07cef3/crates/node/src/add_ons/rollup.rs#L191-L204

PR #151, opened 2025-06-03
feat: add a `finalized` predicate to `L1MessageProvider`

colinlyguo:
# Overview
This PR introduces a `finalized` predicate for `L1MessageProvider` to allow filtering L1 messages based on finalization status instead of block depth. It also adds CLI flags to allow users to choose between `block depth` mode and `finalized` mode for L1 message inclusion.

closes [#143](https://github.com/scroll-tech/rollup-node/issues/143).

### DIFF
--- a/crates/indexer/src/lib.rs
+++ b/crates/indexer/src/lib.rs
@@ -59,6 +59,11 @@ impl<ChainSpec: ScrollHardforks + EthChainSpec + Send + Sync + 'static> Indexer<
         }
     }
 
+    /// Returns the current L1 finalized block number.
+    pub fn get_l1_finalized_block_number(&self) -> u64 {
+        self.l1_finalized_block_number.load(Ordering::Relaxed)
+    }
+
     /// Handles an L2 block.
     pub fn handle_block(
         &mut self,
--- a/crates/node/src/args.rs
+++ b/crates/node/src/args.rs
@@ -110,4 +110,15 @@ pub struct SequencerArgs {
     /// The fee recipient for the sequencer.
     #[arg(long = "sequencer.fee-recipient", id = "sequencer_fee_recipient", value_name = "SEQUENCER_FEE_RECIPIENT", default_value_t = SCROLL_FEE_VAULT_ADDRESS)]
     pub fee_recipient: Address,
+    /// L1 message inclusion mode: either "block-depth" or "finalized"
+    #[arg(
+        long = "sequencer.l1-inclusion-mode",
+        id = "sequencer_l1_inclusion_mode",
+        value_name = "SEQUENCER_L1_INCLUSION_MODE",
+        default_value = "block-depth"
+    )]
+    pub l1_inclusion_mode: String,
+    /// L1 block depth for block depth mode (number of confirmations to wait)
+    #[arg(long = "sequencer.l1-block-depth", id = "sequencer_l1_block_depth", value_name = "SEQUENCER_L1_BLOCK_DEPTH", default_value_t = constants::DEFAULT_L1_BLOCK_DEPTH)]
+    pub l1_block_depth: u64,
 }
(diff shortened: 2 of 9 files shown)


crates/sequencer/src/lib.rs
```diff
@@ -34,8 +46,10 @@ pub struct Sequencer<P> {
     max_l1_messages_per_block: u64,
     /// The current l1 block number.
     l1_block_number: u64,
-    /// The L1 block depth at which L1 messages should be included in the payload.
-    l1_block_depth: u64,
+    /// The L1 finalized block number.
+    l1_finalized_block_number: Arc<AtomicU64>,
```
frisitano, line 50: I think we can use `u64` instead of `Arc<AtomicU64>` here.

crates/node/src/args.rs
```diff
@@ -110,4 +110,15 @@ pub struct SequencerArgs {
...
+    #[arg(
+        long = "sequencer.l1-inclusion-mode",
+        id = "sequencer_l1_inclusion_mode",
+        value_name = "SEQUENCER_L1_INCLUSION_MODE",
+        default_value = "block-depth"
+    )]
+    pub l1_inclusion_mode: String,
+    /// L1 block depth for block depth mode (number of confirmations to wait)
+    #[arg(long = "sequencer.l1-block-depth", id = "sequencer_l1_block_depth", value_name = "SEQUENCER_L1_BLOCK_DEPTH", default_value_t = constants::DEFAULT_L1_BLOCK_DEPTH)]
+    pub l1_block_depth: u64,
```
frisitano, line 123: I think we can combine these into a single argument `sequencer.l1-message-inclusion-mode`. This should be of type `L1MessageInclusionMode` instead of `String`. You can then implement `impl FromStr for L1MessageInclusionMode` which would look something like:

```rust
impl FromStr for L1MessageInclusionMode {
    type Err = String;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if s.eq_ignore_ascii_case("finalized") {
            Ok(L1MessageInclusionMode::Finalized)
        } else if let Some(rest) = s.strip_prefix("depth:") {
            rest.parse::<u64>()
                .map(L1MessageInclusionMode::BlockDepth)
                .map_err(|_| "Expected a number after `depth:`".to_string())
        } else {
            Err("Expected `finalized` or `depth:<number>`".to_string())
        }
    }
}
``` 

You can then modify the argument definition to be something like
```rust
#[arg(
    long = "sequencer.l1-inclusion-mode",
    id = "sequencer_l1_inclusion_mode",
    value_name = "SEQUENCER_L1_INCLUSION_MODE",
    default_value = "finalized",
    value_parser = clap::builder::ValueParser::new(L1MessageInclusionMode::from_str)
)]
```

This allows us to use `--sequencer.l1-inclusion-mode finalized` or `--sequencer.l1-inclusion-mode depth:10`. What do you think?

crates/indexer/src/lib.rs
```diff
@@ -59,6 +59,11 @@ impl<ChainSpec: ScrollHardforks + EthChainSpec + Send + Sync + 'static> Indexer<
         }
     }
 
+    /// Returns the current L1 finalized block number.
+    pub fn get_l1_finalized_block_number(&self) -> u64 {
```
frisitano, line 63: Is this ever used?

frisitano (commented):
Added a few comments inline specifically about unifying the two arguments which have been introduced into a single argument and also doing the parsing at the CLI layer to ensure we catch bad cli arguments early. What do you think?

crates/sequencer/src/lib.rs
colinlyguo: A race condition exists when writing `l1_finalized_block_number` when invoking `set_l1_finalized_block_number` by `monitor` and reading `l1_finalized_block_number` when invoking `build_payload_attributes` by `sequencer`. Do you think this is a valid concern?

PR #153, opened 2025-06-03
feat: metrics

greged93:
Adds metrics to the RN.

Current dashboard view:
Screenshot 2025-06-03 at 18 41 06
Screenshot 2025-06-03 at 18 42 52


greged93:
I think the impact should be limited, we aren't collecting a lot of metrics yet. But I can't really give you a clear answer.

PR #154, opened 2025-06-04
Feat: header metadata

greged93:
Adds fetching of the missing header metadata during migration.

CLI result:
Screenshot 2025-06-04 at 17 43 45


crates/database/db/src/db.rs
```diff
@@ -134,12 +134,13 @@ mod test {
     }
 
     #[tokio::test]
+    #[ignore]
```
greged93, line 137: current test migration doesn't contain any block data, maybe need to add a test migration?

frisitano: I think it's fine to ignore for now.

frisitano (dismissed):
Looks good, added some optional nits inline. LGTM.

crates/database/migration/src/m20250408_150338_load_header_metadata.rs
```diff
@@ -0,0 +1,304 @@
...
+                .with_key("eta", |state: &ProgressState, w: &mut dyn Write| write!(w, "{:.1}s", state.eta().as_secs_f64()).unwrap())
+                .progress_chars("#>-")
+        );
+    pb.set_position(0);
+
+    // init variables.
+    let mut buf = Vec::with_capacity(total_size as usize);
+    let mut index = 0;
+    let mut cursor = 0;
+    let mut tasks = FuturesOrdered::new();
```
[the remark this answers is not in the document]
greged93: I'll add the extra logic, I don't think it's very long, could be nice to track progress better.

crates/database/migration/src/m20250408_150338_load_header_metadata.rs
```diff
@@ -0,0 +1,316 @@
...
+
+            // batch the insertion to avoid `too many SQL variables` error.
+            const MAX_BATCH_SIZE: usize = 3000;
+            let mut cursor = 0;
+            while cursor < records.len() {
+                let start = cursor;
+                let end = (start + MAX_BATCH_SIZE).min(records.len());
+                Entity::insert_many(records[start..end].to_vec()).exec(db).await?;
+                cursor = end;
+            }
```
greged93, line 50: This was needed due to the amount of data we are inserting. I also checked that both Sepolia and Mainnet contained the correct amount of headers and verified the last header for both.

frisitano (approved):
Looks great! LGTM

Issue #163, opened 2025-06-06
[Singer] AWS KMS Signer


frisitano:
# Overview
We should implement an AWS KMS signer such that we never have to have the signing key on the node. This can be achieved using [`alloy-signer-aws`](https://github.com/alloy-rs/alloy/tree/main/crates/signer-aws) crate.

PR #165, opened 2025-06-13
feat: add AWS signer configuration for sequencer signing

colinlyguo:
# Overview
This PR extends the sequencer signing configuration by adding AWS KMS support alongside the existing private key file option. It introduces the `--signer.aws-kms-key-id` parameter and implements mutually exclusive validation between the two signing methods to ensure exactly one is specified when the sequencer is enabled (unless in test mode).

closes https://github.com/scroll-tech/rollup-node/issues/163


README.md
```diff
@@ -122,19 +122,34 @@ To run a sequencer node you should build the binary in release mode using the in
 
 Then, you can run the sequencer node with the following command:
 
+### Using Private Key File
```
frisitano, line 125: Should we mention that the key should be hex encoded here as well?

crates/node/src/add_ons/rollup.rs
```diff
@@ -256,3 +269,25 @@ impl RollupManagerAddOn {
...
+        .map_err(|e| eyre::eyre!("Failed to initialize AWS KMS signer: {}", e))?;
+
+    tracing::info!(
+        "Created AWS KMS signer with address: {} for chain ID: {}",
+        aws_signer.address(),
+        chain_id
+    );
+
+    Ok(aws_signer)
+}
```
frisitano, line 293: Let's refactor this slightly. Instead of having this logic and the above logic on line 220 for creating the signer. We should add a function on `SignerArgs` such that we have something like this:

```rust
impl SignerArgs {
    pub fn signer(&self, chain_id: u64) -> Option<Box<dyn alloy_signer::Signer>> {
        todo!()
    }
}
```

What do you think?

frisitano (changes requested):
Added some minor comments inline about moving the `Signer` creation logic into the `SignerArgs`.

crates/node/src/add_ons/rollup.rs
colinlyguo: Makes sense, which is a good practice to encapsulate the logic. refactored in commit [1da8e80](https://github.com/scroll-tech/rollup-node/pull/165/commits/1da8e80ead62343f7e1dd342201e5f08856baa3b).

frisitano (approved):
looks great! lgtm

crates/node/src/args.rs
```diff
@@ -146,9 +152,83 @@ pub struct SignerArgs {
...
+
+            Ok(Some(Box::new(aws_signer)))
+        } else {
+            // This should not happen because ScrollRollupNodeConfig::validate() should catch this,
+            // but we handle it defensively
+            tracing::error!("No signer configured for chain ID: {}", chain_id);
+            Err(eyre::eyre!(
+                "No signer configured: neither key_file nor aws_kms_key_id is set for chain ID {}",
+                chain_id
+            ))
```
frisitano, line 229: Isn't it possible that we don't have a signer enabled, so we should return `Ok(None)`?

frisitano (changes requested):
Left a comment inline.

PR #173, opened 2025-06-17
feat: break payload building

greged93:
Integrates changes done on Reth related to payload building breaker and adds e2e testing for the breaker.


frisitano (approved):
Looks good! lgtm

Issue #176, opened 2025-06-18
[CI] Revert nextest version to `latest` in CI


frisitano:
# Overview
We had to temporarily pin a `nextest` version in CI due to an issue with the `latest` version. When this is fixed we can revert.
reference: https://github.com/taiki-e/install-action/issues/1005

PR #177, opened 2025-06-18
fix: pin nextest version

frisitano:
# Overview
The latest version of `nextest` introduces a bug when using it in CI. This issue pins the nextest version. I have created an issue to track this in the future #176.


greged93 (approved):
lgtm!

PR #180, opened 2025-06-24
feat: update block data hint database models

greged93:
Updates the database models for the block data hint.


crates/providers/src/execution_payload.rs
```diff
@@ -1,88 +0,0 @@
-use std::fmt::Debug;
```
greged93, line 1: I removed this trait, it wasn't very useful in the end, just a wrapper around `Provider`...

crates/database/migration/src/m20250408_150338_load_header_metadata.rs
```diff
@@ -270,23 +277,46 @@ impl MetadataDecoder {
...
 
-        if buf.len() < B256::len_bytes() + seal_length + 2 {
+        // flag + vanity index + state root + coinbase + nonce + seal
+        let total_expected_size = 2 * size_of::<u8>() +
+            B256::len_bytes() +
+            Address::len_bytes() * has_coinbase as usize +
+            size_of::<u64>() * has_nonce as usize +
+            seal_length;
+
+        if buf.len() < total_expected_size {
```
frisitano, line 293: Should we raise an error in this case?

crates/engine/src/future/mod.rs
```diff
@@ -235,24 +238,27 @@ where
...
-        .execution_payload_for_block((fcs.safe_block_info().number + 1).into())
+    let maybe_execution_payload = provider
+        .get_block((fcs.safe_block_info().number + 1).into())
+        .full()
         .await
         .map_err(|_| EngineDriverError::ExecutionPayloadProviderUnavailable)?
-        .filter(|ep| matching_payloads(&payload_attributes, ep, fcs.safe_block_info().hash));
+        .map(|b| b.into_consensus().map_transactions(|tx| tx.inner.into_inner()))
+        .filter(|b| block_matches_attributes(&payload_attributes, b, fcs.safe_block_info().hash))
+        .map(|b| ExecutionPayload::from_block_slow(&b).0);
```
frisitano, line 261: Do we need to do this final map? Can we just use the block instead for the `if let` statement below?

frisitano (dismissed):
Looks good. Left minor nits inline.

Issue #182, opened 2025-06-26
[Indexer] Refactor `Indexer` to `ChainOrchestrator`


frisitano:
# Overview
This is a proposal to refactor the `Indexer` into a `ChainOrchestrator` component. The `ChainOrchestrator` component will have the following responsibilities:
- Determine when we are in optimistic mode
- Validate that L1 messages are correct and continuous
- Detect reorgs on L2 and issue an appropriate event
- Be the primary driver of the EN

## Optimistic Mode
We define optimistic mode as a mode when we have not fully synced from L1, and as such, we will not be in a position to validate all L1 messages.

When a node starts up, the `ChainOrchestrator` will be instantiated with `optimistic_mode = true`. We then will have to wait for the `L1Watcher` to sync, at which point it should emit a `L1Synced` event. We need to ensure that all previously emitted batches and L1 messages have been consolidated and indexed before switching the `ChainOrchestrator` to `optimistic_mode = false`. At this point we should validate that any blocks that we have optimistically synced have valid L1 messages included in them. Following this, for every new block received over the network, we should validate the L1 messages before pushing the blocks to the EN.

## L2 Reorgs
The `ChainOrchestrator` should be responsible for detecting L2 reorgs. When we receive a block over P2P, we should reconcile the block with the chain data we have in the database. If we have a gap between the database chain tip and the P2P block, then we should recursively request parent blocks until we find the highest common ancestor. If the highest common ancestor is the chain tip, then we know this chain extends the chain the RN is following; if it is some deeper block, then we have detected a reorg and must unwind the chain to the highest common ancestor. We must then validate L1 messages from the new chain blocks and issue messages for the EN to execute these blocks and issue an FCU. On startup, we encounter a challenge in which the gap between the indexed data in the database and the block received over P2P is very large. As such, we need a means of addressing this. A simple approach would be to optimistically sync the last N (~300?) blocks and persist them in the database such that on future blocks we can detect reorgs against this optimistic chain. A more robust solution would be to fetch the latest finalized block from L1 and then use that as an anchor to recursively sync back to. In the name of practicality, I propose we implement the former solution first and then subsequently implement the latter.

PR #185, opened 2025-07-02
feat: chain orchestrator

frisitano:
closes: #182
closes https://github.com/scroll-tech/reth/issues/243

### DIFF
--- a/crates/database/migration/src/lib.rs
+++ b/crates/database/migration/src/lib.rs
@@ -6,6 +6,7 @@ mod m20250408_132123_add_header_metadata;
 mod m20250408_150338_load_header_metadata;
 mod m20250411_072004_add_l2_block;
 mod m20250616_223947_add_metadata;
+mod m20250825_093350_remove_unsafe_l2_blocks;
 mod migration_info;
 pub use migration_info::{
     MigrationInfo, ScrollDevMigrationInfo, ScrollMainnetMigrationInfo, ScrollSepoliaMigrationInfo,
@@ -23,6 +24,7 @@ impl<MI: MigrationInfo + Send + Sync + 'static> MigratorTrait for Migrator<MI> {
             Box::new(m20250408_150338_load_header_metadata::Migration::<MI>(Default::default())),
             Box::new(m20250411_072004_add_l2_block::Migration::<MI>(Default::default())),
             Box::new(m20250616_223947_add_metadata::Migration),
+            Box::new(m20250825_093350_remove_unsafe_l2_blocks::Migration),
         ]
     }
 }
--- a/crates/database/migration/src/m20250825_093350_remove_unsafe_l2_blocks.rs
+++ b/crates/database/migration/src/m20250825_093350_remove_unsafe_l2_blocks.rs
@@ -1,0 +1,28 @@
+use sea_orm::Statement;
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute(Statement::from_sql_and_values(
+                manager.get_database_backend(),
+                r#"
+        DELETE FROM l2_block
+        WHERE batch_index IS NULL
+        "#,
+                [],
+            ))
+            .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, _manager: &SchemaManager) -> Result<(), DbErr> {
+        Ok(())
+    }
+}


greged93 (commented):
Couple of comments and some small nits and leftover code to clean.

crates/database/db/src/operations.rs
```diff
@@ -382,33 +421,49 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
-                        models::l2_block::Column::BlockHash,
-                        models::l2_block::Column::BatchHash,
-                        models::l2_block::Column::BatchIndex,
-                    ])
-                    .to_owned(),
-            )
-            .exec(self.get_connection())
-            .await?;
+        // We only insert safe blocks into the database, we do not persist unsafe blocks.
+        if batch_info.is_some() {
```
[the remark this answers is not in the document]
frisitano: So currently, we use this function in 2 circumstances: 

1. Inserting blocks from L1 consolidation
2. For updating the L2 block number at which a L1 message has been included (in this case, we do not want to persist the block itself, only map the L1 message to the L2 block number).

crates/database/migration/src/m20220101_000001_create_batch_commit_table.rs
```diff
@@ -23,7 +24,29 @@ impl MigrationTrait for Migration {
...
+        "#,
+                vec![
+                    0u64.into(),
+                    vec![0u8; HASH_LENGTH as usize].into(),
+                    0u64.into(),
+                    0u64.into(),
+                    vec![].into(),
+                    None::<Vec<u8>>.into(),
+                    0u64.into(),
+                ],
```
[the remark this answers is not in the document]
frisitano: This is used to mark the genesis block stored in the `l2_block` table as safe. The genesis block in that table can only be marked as safe if it has a batch that it can be related to. We need this to be safe as when the node is started the `ChainOrchestrator` loads the latest safe chain into the in-memory chain.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+            // If we are in optimistic mode and the received chain can not be reconciled with the
+            // in-memory chain we break. We will reconcile after optimistic sync has completed.
+            if *optimistic_mode.lock().await &&
+                received_chain_headers.last().unwrap().number <
+                    current_chain_headers.front().unwrap().number
+            {
+                return Ok(ChainOrchestratorEvent::InsufficientDataForReceivedBlock(
+                    received_block.hash_slow(),
+                ))
+            }
```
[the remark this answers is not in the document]
frisitano: We may start with `received_chain_headers.last().number > current_chain_headers.front().number` but as we fetch more more headers for `received_chain_headers` via the loop we can reach a point where `received_chain_headers.last().number < current_chain_headers.front().number` as and such we should terminate the loop. Let me know if this makes sense to you.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+                .await?
+                .into_data()
+            {
+                received_chain_headers.push(header.clone());
+                received_header_tail = header;
+            } else {
+                return Err(ChainOrchestratorError::MissingBlockHeader {
+                    hash: received_header_tail.parent_hash,
+                });
+            }
```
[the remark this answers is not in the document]
frisitano: Because we haven't been able to reconcile the `received_chain_headers` with `current_chain_headers` which suggests that the fork is deeper than the current block number of `received_chain_headers`.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+                    new_blocks, peer_id, signature,
+                )))
+            }
+            // If we are re-organizing the in-memory chain, we need to split the chain at the reorg
+            // point and extend it with the new blocks.
+            position => {
+                // reorg the in-memory chain to the new chain and issue a reorg event.
+                let mut new_chain = Chain::new(CHAIN_BUFFER_SIZE);
+                new_chain.extend(current_chain_headers.iter().take(position).cloned());
+                new_chain.extend(received_chain_headers);
```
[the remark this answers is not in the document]
frisitano: can you elaborate on why you think this is the case please?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+                            hash: current_chain_headers.front().unwrap().parent_hash,
+                        });
+                    }
+                }
+            }
+
+            // We search the in-memory chain to see if we can reconcile the block import.
+            if let Some(pos) = current_chain_headers
+                .iter()
+                .rposition(|h| h.hash_slow() == received_header_tail.parent_hash)
```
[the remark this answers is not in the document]
frisitano: I think there is a bug here, we shouldn't be updating `received_header_tail = header`. Let me update this and add some test cases for deep reorgs.

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
greged93: In the current implementation I did, instead of extending the bounded vec and dropping hold values, I drop the iterator provided in extend until it fits in the vec. This seems incorrect to me.

crates/watcher/src/lib.rs
```diff
@@ -212,8 +220,17 @@ where
...
                 tokio::time::sleep(SLOW_SYNC_INTERVAL).await;
+            } else if self.current_block_number == self.l1_state.head {
+                // if we have synced to the head of the L1, notify the channel and set the
+                // `is_synced`` flag.
+                if let Err(L1WatcherError::SendError(_)) = self.notify(L1Notification::Synced).await
+                {
+                    tracing::warn!(target: "scroll::watcher", "L1 watcher channel closed, stopping the watcher");
+                    break;
+                }
+                self.is_synced = true;
```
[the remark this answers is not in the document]
greged93: hmmm but so if we lose a provider for 12 minutes we might enter an edge case we can't exit from?

crates/database/db/src/operations.rs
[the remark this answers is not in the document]
greged93: Also, based on the comment, should we update the database model to reflect that every l2 block in database should have a batch?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+                received_block.header.hash_slow(),
+            ));
+        };
+
+        // We fetch headers for the received chain until we can reconcile it with the chain we
+        // currently have in-memory.
+        let mut received_chain_headers = vec![received_block.header.clone()];
+        let mut received_header_tail = received_block.header.clone();
+
+        // We should never have a re-org that is deeper than the current safe head.
```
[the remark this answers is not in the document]
frisitano: I think this is worthy of open discourse amongst the team. In my opinion, safe should mean that it is highly probable that the block will be settled at some later time. The only way a block associated with a batch posted to L1 would not be settled is if the batch was reorged (and not included later), and given that reorgs are relatively infrequent and shallow, I think the current approach is reasonable. Furthermore, I think we can consider a block as finalized if the batch associated with the block is included in a finalzied block. Provided we work under the assumption of completeness of the proof system and permissionless submission of proofs then I think we can mark these blocks as described before any proofs are posted to L1. Lets open this discussion for the team more broadly.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1520 @@
...
+                received_chain_headers.push(header.clone());
+            } else {
+                return Err(ChainOrchestratorError::MissingBlockHeader {
+                    hash: received_chain_headers.last().unwrap().parent_hash,
+                });
+            }
+        };
+
+        // Reverse the new chain headers to have them in the correct order.
+        received_chain_headers.reverse();
```
greged93, line 330: Personally think we would gain in code clarity if received_chain_headers and current_chain_headers were ordered in the same way.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1520 @@
...
+        let chain_buffer_size = self.chain_buffer_size;
+        let fut =
+            self.handle_metered(
+                ChainOrchestratorItem::InsertL2Block,
+                Box::pin(async move {
+                    // If we are in optimistic mode and the L1 is synced, we consolidate the chain
+                    // and disable optimistic mode to enter consolidated mode
+                    // (consolidated_mode = !optimistic_mode).
+                    let consolidated = if !*optimistic_mode.lock().await {
+                        true
```
greged93, line 394: we don't perform any consolidation here. Is this because in `!optimistic_mode`, the consolidation happens in `handle_new_block`?

crates/engine/src/driver.rs
```diff
@@ -122,17 +122,24 @@ where
...
+        self.chain_imports.push_back(chain_import);
+        self.waker.wake();
+    }
+
+    /// Optimistically syncs the chain to the provided block info.
+    pub fn handle_optimistic_sync(&mut self, block_info: BlockInfo) {
+        tracing::info!(target: "scroll::engine", ?block_info, "optimistic sync request received");
+
+        // Purge all pending block imports.
+        self.chain_imports.clear();
```
greged93, line 137: why do we purge all pending block imports?

crates/network/src/manager.rs
```diff
@@ -182,9 +217,45 @@ impl<N: FullNetwork> ScrollNetworkManager<N> {
             }
         }
     }
+
+    /// Handles a new block received from the eth-wire protocol.
+    fn handle_eth_wire_block(
```
greged93, line 222: makes sense to have this here 👌

crates/network/src/manager.rs
```diff
@@ -211,6 +282,15 @@ impl<N: FullNetwork> Stream for ScrollNetworkManager<N> {
             return Poll::Ready(Some(this.on_scroll_wire_event(event)));
         }
 
+        // Handle blocks received from the eth-wire protocol.
+        while let Some(Poll::Ready(Some(block))) =
```
greged93, line 286: note (not relevant for now but lets keep it in mind): at some point we might want to implement the same [logic](https://github.com/scroll-tech/reth/blob/50c207d62e2ae3236e9d3ca0aab23601e5dee028/crates/net/network/src/manager.rs#L1126) as in Reth to avoid having some components in the node monopolizing all resources.

crates/sequencer/tests/e2e.rs
```diff
@@ -599,7 +598,7 @@ async fn can_build_blocks_and_exit_at_gas_limit() {
     let chain_spec = SCROLL_DEV.clone();
     const MIN_TRANSACTION_GAS_COST: u64 = 21_000;
     const BLOCK_BUILDING_DURATION: Duration = Duration::from_millis(250);
-    const BLOCK_GAP_TRIGGER: u64 = 100;
+    // const BLOCK_GAP_TRIGGER: u64 = 100;
```
greged93, line 601: remove?

greged93 (commented):
couple of extra comments and questions. Also I see we have a lot more `unwrap`s in the code, are all these safe to keep?

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
frisitano: updated and test added, please re-review

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
frisitano: updated

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1502 @@
...
+    use std::{collections::HashMap, ops::RangeInclusive, sync::Arc};
+
+    type ScrollBody = <ScrollBlock as Block>::Body;
+
+    const TEST_OPTIMISTIC_SYNC_THRESHOLD: u64 = 100;
+
+    /// A headers+bodies client that stores the headers and bodies in memory, with an artificial
+    /// soft bodies response limit that is set to 20 by default.
+    ///
+    /// This full block client can be [Clone]d and shared between multiple tasks.
```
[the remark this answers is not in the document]
frisitano: no it's not because `Clone` is the trait that is implemented, and this pattern makes `[Clone]` linkable in rust docs

greged93 (commented):
reviewing l2geth code, I noticed we might be missing checks on L1 messages.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1564 @@
...
+
+    blocks
+}
+
+/// Validates the L1 messages in the provided blocks against the expected L1 messages synced from
+/// L1.
+async fn validate_l1_messages(
+    blocks: &[ScrollBlock],
+    database: Arc<Database>,
+) -> Result<(), ChainOrchestratorError> {
```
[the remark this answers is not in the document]
greged93: > Or are you suggesting we do this in the consensus on reth (as I believe you have already in the open PR)?

I'm actually not sure if we should have these checks in Reth and/or Rollup node. What do you think?

> Regarding the start index, how about we introduce an l1_message_index: Arc<Mutex<IndexType> in memory which is used to determine the cursor start.

Won't we have an issue with L1 reorgs with this?

crates/network/src/manager.rs
frisitano: Completely agree. lets add an issue for this.

crates/chain-orchestrator/src/lib.rs
greged93: works for me, should we raise an issue with all improvements listed?

crates/engine/src/driver.rs
greged93: should we also clear the pending futures then?

crates/watcher/src/lib.rs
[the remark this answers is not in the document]
greged93: Should we do this in another PR?

greged93 (approved):
just need to resolve the pending comments and open 1 issues related to L1 watcher sync.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+                        .number
+                        .saturating_sub(1) <=
+                        latest_safe_block.number
+                    {
+                        tracing::info!(target: "scroll::chain_orchestrator", hash = %latest_safe_block.hash, number = %latest_safe_block.number, "reached safe block number for current chain - terminating fetching.");
+                        break;
+                    }
+                    tracing::trace!(target: "scroll::chain_orchestrator", number = ?(current_chain_headers.front().expect("chain can not be empty").number - 1), "fetching block for current chain");
+                    if let Some(block) = l2_client
+                        .get_block_by_hash(
```
jonastheis, line 320: What's the timeout for this operation? What if none of our neighbors supplies this block?

crates/node/src/constants.rs
```diff
@@ -19,7 +19,10 @@ pub(crate) const DEFAULT_PAYLOAD_BUILDING_DURATION: u64 = 800;
 pub(crate) const DEFAULT_PAYLOAD_SIZE_LIMIT: u64 = 122_880;
 
 /// The gap in blocks between the P2P and EN which triggers sync.
-pub(crate) const BLOCK_GAP_TRIGGER: u64 = 500_000;
+pub(crate) const BLOCK_GAP_TRIGGER: u64 = 100_000;
+
+/// The number of block headers to keep in the in-memory chain buffer in the chain orchestrator.
+pub(crate) const CHAIN_BUFFER_SIZE: usize = 2000;
```
jonastheis, line 25: Wondering if 2000 is enough? How much memory would this occupy?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+                    }
+                }
+            }
+
+            // We search the in-memory chain to see if we can reconcile the block import.
+            if let Some(pos) = current_chain_headers.iter().rposition(|h| {
+                h.hash_slow() ==
+                    received_chain_headers.last().expect("chain can not be empty").parent_hash
+            }) {
+                // If the received fork is older than the current chain, we return an event
```
jonastheis, line 346: How do we know here that this is a fork/reorg?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+                        peer_id,
+                        signature,
+                    });
+                }
+                break pos;
+            }
+
+            // If the current header block number is less than the latest safe block number then
+            // we should error.
+            if received_chain_headers.last().expect("chain can not be empty").number <=
```
jonastheis, line 366: Shouldn't we do this quick check already earlier?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+                latest_safe_block.number
+            {
+                return Err(ChainOrchestratorError::L2SafeBlockReorgDetected);
+            }
+
+            tracing::trace!(target: "scroll::chain_orchestrator", number = ?(received_chain_headers.last().expect("chain can not be empty").number - 1), "fetching block");
```
jonastheis, line 372: can't we store `received_chain_headers.last()` in a variable somewhere to avoid calling the expect dozens of times?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+
+            // We search the in-memory chain to see if we can reconcile the block import.
+            if let Some(pos) = current_chain_headers.iter().rposition(|h| {
+                h.hash_slow() ==
+                    received_chain_headers.last().expect("chain can not be empty").parent_hash
+            }) {
+                // If the received fork is older than the current chain, we return an event
+                // indicating that we have received an old fork.
+                if (pos < current_chain_headers.len() - 1) &&
+                    current_chain_headers.get(pos + 1).expect("chain can not be empty").timestamp >
```
jonastheis, line 349: This seems like quite important: we don't react to reorgs based on timestamp of a block. what if the sequencer's time is faulty?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+    pub optimistic_mode: Arc<Mutex<bool>>,
+    pub optimistic_sync_threshold: u64,
+    pub network_client: Arc<BC>,
+    pub database: Arc<Database>,
+    pub chain_buffer_size: usize,
+}
+
+/// Consolidates the chain by reconciling the in-memory chain with the L2 client and database.
+/// This is used to ensure that the in-memory chain is consistent with the L2 chain.
+async fn consolidate_chain<P: Provider<Scroll> + 'static>(
```
jonastheis, line 815: Why is it enough to consolidate only back to the safe head? Is it because the safe block is derived from L1 and the L1 messages until that point are verified through the L1 derivation pipeline?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+        chain_spec: Arc<ChainSpec>,
+        l1_message: TxL1Message,
+        l1_block_number: u64,
+        block_timestamp: u64,
+    ) -> Result<Option<ChainOrchestratorEvent>, ChainOrchestratorError> {
+        let event = ChainOrchestratorEvent::L1MessageCommitted(l1_message.queue_index);
+
+        let queue_hash = if chain_spec
+            .scroll_fork_activation(ScrollHardfork::EuclidV2)
+            .active_at_timestamp_or_number(block_timestamp, l1_block_number) &&
```
jonastheis, line 646: Why pass `l1_block_number` here?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+            txn.delete_l2_blocks_gt_batch_index(prev_batch_index).await?;
+            txn.get_highest_block_for_batch_index(prev_batch_index).await?
+        } else {
+            None
+        };
+
+        let event = ChainOrchestratorEvent::BatchCommitIndexed {
+            batch_info: BatchInfo::new(batch.index, batch.hash),
+            l1_block_number: batch.block_number,
+            safe_head: new_safe_head,
```
jonastheis, line 690: Why is the `safe_head` only part of this event in case there was a batch revert? Couldn't it always be part of the event and have a separate field if it was indeed a batch revert (if we even need that information)?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+            finalized_block =
+                Self::fetch_highest_finalized_block(database, batch_hash, l2_block_number).await?;
+        }
+
+        let event = ChainOrchestratorEvent::BatchFinalized(batch_hash, finalized_block);
+        Ok(Some(event))
+    }
+
+    /// Returns the highest finalized block for the provided batch hash. Will return [`None`] if the
+    /// block number has already been seen by the indexer.
```
jonastheis, line 725: Let's also update the comments where `indexer` -> `chain orchestrator`

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
jonastheis: We should replicate the l2geth implementation as close as possible. This also means the order within the block etc. We should open a separate issue to revisit this after the PR is merged.

crates/watcher/src/lib.rs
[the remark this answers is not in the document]
frisitano: @jonastheis is it fair to say this is captured in this issue https://github.com/scroll-tech/rollup-node/issues/252?

crates/engine/src/driver.rs
frisitano: I presume you are referring to `engine_future`? If so, we could do but this engine future may be related to an L1 consolidation, in which case we wouldn't want to clear it. I think it's fine to leave as unless I'm misssing something?

crates/node/src/args.rs
```diff
@@ -394,26 +418,49 @@ impl Default for ConsensusAlgorithm {
...
+    }
+}
+
 /// The network arguments.
 #[derive(Debug, Clone, clap::Args)]
 pub struct NetworkArgs {
     /// A bool to represent if new blocks should be bridged from the eth wire protocol to the
     /// scroll wire protocol.
-    #[arg(long = "network.bridge", default_value_t = true)]
+    #[arg(long = "network.bridge")]
```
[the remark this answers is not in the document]
frisitano: Currently, there is no way to set this to false via CLI. With booleans, the absence of the argument represents false. However, because we are setting `default_value_t = true`, this remains true even in the absence of the CLI argument.

crates/node/src/args.rs
```diff
@@ -422,11 +469,19 @@ pub struct NetworkArgs {
         value_name = "NETWORK_SEQUENCER_URL"
     )]
     pub sequencer_url: Option<String>,
+    /// A bool that represents if blocks should be gossiped over the eth-wire protocol.
+    #[arg(long = "network.eth-wire-gossip")]
```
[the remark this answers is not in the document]
frisitano: Your description of the interaction is correct. 

The intention of this flag was to mitigate duplication of block announcements. However, I think a more robust solution would be to handle it on the receiving end as described in your issue #251. I will revert this change.

crates/network/src/manager.rs
```diff
@@ -191,9 +236,53 @@ impl<N: FullNetwork<Primitives = ScrollNetworkPrimitives>> ScrollNetworkManager<
...
+        // from the last `ECDSA_SIGNATURE_LEN` bytes of the extra data field as specified by
+        // the protocol.
+        if let Some(signature) = extra_data
+            .len()
+            .checked_sub(ECDSA_SIGNATURE_LEN)
+            .and_then(|i| Signature::from_raw(&extra_data[i..]).ok())
+        {
+            Some(NetworkManagerEvent::NewBlock(NewBlockWithPeer { peer_id, block, signature }))
+        } else {
+            tracing::warn!(target: "scroll::bridge::import", peer_id = %peer_id, "Failed to extract signature from block extra data");
```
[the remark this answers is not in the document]
frisitano: Good catch yes we should. will update.

crates/database/db/src/models/l2_block.rs
```diff
@@ -9,21 +9,17 @@ pub struct Model {
     #[sea_orm(primary_key)]
     block_number: i64,
     block_hash: Vec<u8>,
-    batch_index: Option<i64>,
-    batch_hash: Option<Vec<u8>>,
+    batch_index: i64,
```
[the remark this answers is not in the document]
frisitano: We don't persist unsafe L2 blocks in the database anymore. They are already persisted in the execution node.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -0,0 +1,1610 @@
...
+            let mut optimistic_headers = Chain::new(chain_buffer_size);
+            optimistic_headers.push_front(received_block.header.clone());
+            while optimistic_headers.len() < chain_buffer_size &&
+                optimistic_headers.first().expect("chain can not be empty").number != 0
+            {
+                tracing::trace!(target: "scroll::chain_orchestrator", number = ?(optimistic_headers.first().expect("chain can not be empty").number - 1), "fetching block");
+                let parent_hash =
+                    optimistic_headers.first().expect("chain can not be empty").parent_hash;
+                let header = network_client
+                    .get_header(BlockHashOrNumber::Hash(parent_hash))
```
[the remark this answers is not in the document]
frisitano: > Is this efficient? Shouldn't we use GetBlockHeaders for this?
Yes, a more efficient means would be to use `GetBlockHeaders`. 

> We could also wait for the EN to finish the sync and then load from our EN? 
I think this is the best solution.

I will add both of these points to the refactoring issue that I create shortly.

crates/node/src/constants.rs
frisitano: I think we can increase this considerably if required. The general idea is that we should keep enough blocks in memory to support reorgs. I was thinking that a reorg of > 2000 blocks is unlikely.

crates/chain-orchestrator/src/lib.rs
frisitano: Because the received chain does not extend the current chain head (`pos < current_chain_headers.len() - 1`) it extends from a deeper block.

crates/chain-orchestrator/src/lib.rs
frisitano: yes, I've moved the check higher.

crates/chain-orchestrator/src/lib.rs
frisitano: correct

crates/chain-orchestrator/src/lib.rs
frisitano: None the less I agree with the general point and I think we should open an issue for this.

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
frisitano: I believe most of these checks have been implemented in the conensus logic of scroll reth https://github.com/scroll-tech/reth/pull/314/files#diff-a12a5304a4193ceb7da625fa5b565eedd96402e98dab5f7eda45db542bf5795dR353. The outstanding check is to ensure that we maintain continuity across blocks in the chain orchestrator. I have opened this issue to track https://github.com/scroll-tech/rollup-node/issues/245

crates/chain-orchestrator/src/event.rs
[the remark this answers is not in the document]
frisitano: Could you provide a bit more content on specific events that are inconsistent / poorly defined?

crates/network/src/manager.rs
```diff
@@ -161,15 +161,13 @@ impl<
...
-        if self.eth_wire_gossip {
-            let eth_wire_new_block = {
-                let td = U128::from_limbs([0, block.block.header.number]);
-                let mut eth_wire_block = block.block.clone();
-                eth_wire_block.header.extra_data = block.signature.clone().into();
-                EthWireNewBlock { block: eth_wire_block, td }
-            };
-            self.inner_network_handle().eth_wire_announce_block(eth_wire_new_block, hash);
-        }
+        let eth_wire_new_block = {
```
jonastheis, line 164: why do we actually need to compute the total difficulty here? isn't it part the block header already? `eth_wire_block.header.difficulty` or is this a different difficulty?

crates/chain-orchestrator/src/lib.rs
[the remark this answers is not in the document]
jonastheis: Ok, we can resolve this convo from my pov

crates/chain-orchestrator/src/lib.rs
jonastheis: Not sure for now it's okay, but we should discuss this. Maybe best to capture in an issue so we can revisit if necessary and don't forget

crates/network/src/manager.rs
frisitano: I believe that's the block difficulty and not the total difficulty?

crates/watcher/src/lib.rs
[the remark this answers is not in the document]
jonastheis: Maybe you could add some more context to this issue what exactly is the current behavior in this PR and what would be the desired behavior.

jonastheis (dismissed):

frisitano:
Tests are passing fine locally - I suspect there is some form of content in CI. Will merge and fix later.

PR #186, opened 2025-07-03
feat: skipped l1 messages bitmap

greged93:
Makes use of the skipped L1 messages bitmap provided during the batch commit call in the derivation pipeline.


frisitano (approved):
LGTM

PR #189, opened 2025-07-03
feat: support directly forward transactions to sequencer

yiweichi:
Counterpart of l2geth https://github.com/scroll-tech/go-ethereum/pull/1208.
Related Reth PR: https://github.com/scroll-tech/reth/pull/265

Added a new flags on NetworkArgs.
`network.sequencer-url`: This will be passed to `ScrollEthApiBuilder`,  If not None, Reth will forward transactions to sequencer's RPC directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,37 +133,37 @@ alloy-signer-local = { version = "1.0.13", default-features = false }
 alloy-transport = { version = "1.0.13", default-features = false }
 
 # scroll-alloy
-scroll-alloy-consensus = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-scroll-alloy-hardforks = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-scroll-alloy-network = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-scroll-alloy-provider = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-scroll-alloy-rpc-types-engine = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
+scroll-alloy-consensus = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
+scroll-alloy-hardforks = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
+scroll-alloy-network = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
+scroll-alloy-provider = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
+scroll-alloy-rpc-types-engine = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
 
 # reth
-reth-chainspec = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-reth-e2e-test-utils = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer" }
-reth-eth-wire-types = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-reth-network = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
-reth-network-api = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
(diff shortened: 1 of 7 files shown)


crates/node/src/test_utils.rs
```diff
@@ -130,10 +130,7 @@ pub async fn generate_tx(wallet: Arc<Mutex<Wallet>>) -> Bytes {
 pub fn default_test_scroll_rollup_node_config() -> ScrollRollupNodeConfig {
     ScrollRollupNodeConfig {
         test: true,
-        network_args: crate::args::NetworkArgs {
-            enable_eth_scroll_wire_bridge: true,
-            enable_scroll_wire: true,
-        },
+        network_args: crate::args::NetworkArgs::default(),
```
frisitano, line 133: I think `crate::args::NetworkArgs::default()` will result in:
```rust
NetworkArgs {
            enable_eth_scroll_wire_bridge: false,
            enable_scroll_wire: false,
        }
```

This is because the default value for a bool is false. You would need to provide a custom `impl Default for NetworkArgs` to modify the default value to make this change valid. 

I think the `default_value_t` will only be applied when parsing the struct using clap and not when using the `::default()` associated method.

```rust
    /// A bool that represents if the scroll wire protocol should be enabled.
    #[arg(long = "network.scroll-wire", default_value_t = true)]
    pub enable_scroll_wire: bool,
```

frisitano (changes requested):
Added comment inline.

frisitano (commented):
Can we add an e2e test for this using the test framework. See example provided:
https://github.com/scroll-tech/rollup-node/blob/1ffce347f344fbae8466c9dc26d1aba8ce6ead8c/crates/node/tests/e2e.rs#L94-L95

yiweichi:
> Can we add an e2e test for this using the test framework. See example provided:
> 
> https://github.com/scroll-tech/rollup-node/blob/1ffce347f344fbae8466c9dc26d1aba8ce6ead8c/crates/node/tests/e2e.rs#L94-L95

Sure, let me do that.

frisitano:
> Sure, let me do that.

Thanks, I think you'll need to set `no_local_transactions_propagation` to true in the node config when launching the test nodes:

https://github.com/scroll-tech/reth/blob/766dd271b0bd71315cef665d1eeb81ab86b891e0/crates/node/core/src/args/txpool.rs#L90-L92

crates/node/tests/e2e.rs
```diff
@@ -164,6 +166,190 @@ async fn can_sequence_and_gossip_blocks() {
...
+
+    // create 2 nodes
+    let mut sequencer_node_config = default_sequencer_test_scroll_rollup_node_config();
+    sequencer_node_config.sequencer_args.block_time = 0;
+    let mut follower_node_config = default_test_scroll_rollup_node_config();
+
+    // Create the chain spec for scroll mainnet with Euclid v2 activated and a test genesis.
+    let chain_spec = (*SCROLL_DEV).clone();
+    let (mut sequencer_node, _tasks, _) =
+        setup_engine(sequencer_node_config, 1, chain_spec.clone(), false, false).await.unwrap();
```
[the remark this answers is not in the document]
yiweichi: Yeah, this won't have any functional implications since we send transaction on follower node side.

But let me set `no_local_transactions_propagation` to true for sequencer as well, since we probably will also do that for mainnet in the future if we want to enable private transaction pool.

Cargo.toml
```diff
@@ -133,37 +133,37 @@ alloy-signer-local = { version = "1.0.13", default-features = false }
...
-reth-payload-primitives = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-primitives = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-primitives-traits = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-provider = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-rpc-builder = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-rpc-server-types = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-tasks = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-tokio-util = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
-reth-tracing = { git = "https://github.com/scroll-tech/reth.git", default-features = false }
+reth-chainspec = { git = "https://github.com/scroll-tech/reth.git", branch = "feat-forward-tx-to-sequencer", default-features = false }
```
georgehao, line 143: why don't merge `feat-forward-tx-to-sequencer` to main?

yiweichi: the branch `feat-forward-tx-to-sequencer` on Reth hasn't got merged, temporarily set to this to pass the CI.

frisitano:
Please remove `branch = "feat-forward-tx-to-sequencer",` from `Cargo.toml` files as we have now merged `reth` PR into default branch (`scroll`).

frisitano (approved):
lgtm

PR #191, opened 2025-07-09
chore: revm v78 update

frisitano:
# Overview
Merge revm-v78 upstream changes.


frisitano:
Accidentally pushed directly to main. No need to revert the commit as it passes and is green - https://github.com/scroll-tech/rollup-node/commit/62997c128c3dc2e82da37b2114965fe558211362. Repo owner should review the repository rules to ensure we can't push to main via SSH git cli but still allow bypass in web app. cc @Thegaram

PR #221, opened 2025-07-27
fix: update reth to add block payload size validation

yiweichi:
This PR is for Issue: https://github.com/scroll-tech/reth/issues/281

Related PR: https://github.com/scroll-tech/reth/pull/287
The Reth repo PR adds a new limit block_da_size_limit when building a new block. Make sure we won't build a block that's too large to be able to put into a Ethereum blob.


crates/node/src/constants.rs
```diff
@@ -13,6 +13,9 @@ pub(crate) const DEFAULT_BLOCK_TIME: u64 = 1000;
 /// The default payload building duration in milliseconds for the sequencer.
 pub(crate) const DEFAULT_PAYLOAD_BUILDING_DURATION: u64 = 800;
 
+/// The default payload building duration in milliseconds for the sequencer.
```
frisitano, line 16: The comment should be updated to reflec the variable

frisitano (changes requested):
Looks good, left one comment inline.

frisitano (approved):
merge reth branch `feat-add-max-da-block-size`, repoint dep and lgtm

PR #226, opened 2025-07-31
feat: log ScrollChainConfig on node startup

### LINKED ISSUE #222: [Bin] Log `ScrollChainConfig` on node start up
Log the chain config on start up: 
https://github.com/scroll-tech/rollup-node/blob/53b3d1cbdcd029f88cff03fee7dda84f2593776b/crates/node/src/main.rs#L25-L41

haazdev:
## Description
Adds logging of the ScrollChainConfig during node launch as requested in issue #222.

## Changes
- Added logging inside the `launch_with_fn` closure in `main.rs`
- Logs the complete chain specification using debug formatting
- Provides operators visibility into chain configuration at startup

## Testing
- Code compiles successfully
- Clippy passes with no warnings
- Formatting is correct

Closes #222

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -23,6 +23,9 @@ fn main() {
             let handle = builder
                 .node(ScrollRollupNode::new(args))
                 .launch_with_fn(|builder| {
+                    // Log ScrollChainConfig during node launch as requested in the issue
+                    let chain = &builder.config().chain;
+                    info!(target: "reth::cli", "ScrollChainConfig: {:?}", chain);
                     // We must use `always_process_payload_attributes_on_canonical_head` in order to
                     // be able to build payloads with the forkchoice state API
                     // on top of heads part of the canonical state. Not


crates/node/src/main.rs
```diff
@@ -23,6 +23,9 @@ fn main() {
             let handle = builder
                 .node(ScrollRollupNode::new(args))
                 .launch_with_fn(|builder| {
+                    // Log ScrollChainConfig during node launch as requested in the issue
+                    let chain = &builder.config().chain;
+                    info!(target: "reth::cli", "ScrollChainConfig: {:?}", chain);
```
greged93, line 28: No need for the comment, can you make the suggested changes and move this after line 44 so it gets logged after launch?
```suggestion
                    let config = &builder.config().chain.config;
                    info!(target: "reth::cli", ?config, "Running with config");
```

greged93 (commented):
thanks for the contribution! just a comment, please fix and then lets merge it

frisitano:
superseeded by #231

PR #231, opened 2025-08-04
feat(consensus): Add consensus check for block sequencing

frisitano:
# Overview
This PR introduces a `should_sequence_block` on the `Consensus` trait such that the manager can asses if the sequencer is authorized to sign a new block for a specific slot.

### DIFF
--- a/crates/node/src/args.rs
+++ b/crates/node/src/args.rs
@@ -137,6 +137,10 @@ impl ScrollRollupNodeConfig {
         RollupManagerHandle,
         Option<Sender<Arc<L1Notification>>>,
     )> {
+        tracing::info!(target: "rollup_node::args",
+            "Building rollup node with config:\n{:#?}",
+            self
+        );
         // Instantiate the network manager
         let scroll_network_manager = ScrollNetworkManager::from_parts(network.clone(), events);
 
@@ -223,24 +227,16 @@ impl ScrollRollupNodeConfig {
         );
 
         // Create the consensus.
-        let consensus: Box<dyn Consensus> = match self.consensus_args.algorithm {
-            ConsensusAlgorithm::Noop => Box::new(NoopConsensus::default()),
-            ConsensusAlgorithm::SystemContract => {
-                let authorized_signer = if let Some(address) = self.consensus_args.authorized_signer
-                {
-                    address
-                } else if let Some(provider) = l1_provider.as_ref() {
-                    provider
-                        .authorized_signer(node_config.address_book.system_contract_address)
-                        .await?
-                } else {
-                    return Err(eyre::eyre!(
-                        "System contract consensus requires either an authorized signer or a L1 provider URL"
-                    ));
-                };
-                Box::new(SystemContractConsensus::new(authorized_signer))
-            }
+        let authorized_signer = if let Some(provider) = l1_provider.as_ref() {
+            Some(
+                provider
+                    .authorized_signer(node_config.address_book.system_contract_address)
+                    .await?,
+            )
+        } else {
+            None
         };
+        let consensus = self.consensus_args.consensus(authorized_signer)?;
 
         let (l1_notification_tx, l1_notification_rx): (Option<Sender<Arc<L1Notification>>>, _) =
             if let Some(provider) = l1_provider.filter(|_| !self.test) {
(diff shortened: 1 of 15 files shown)


colinlyguo (approved):
LGTM. After this PR, is an L1 node (with system contract slot correctly initialized) needed for local running?

crates/node/src/args.rs
```diff
@@ -297,6 +328,45 @@ pub struct DatabaseArgs {
...
+    #[arg(
+        long = "consensus.algorithm",
+        value_name = "CONSENSUS_ALGORITHM",
+        default_value = "system-contract"
+    )]
+    pub algorithm: ConsensusAlgorithm,
+
+    /// The optional authorized signer for system contract consensus.
+    #[arg(long = "consensus.authorized-signer", value_name = "ADDRESS")]
+    pub authorized_signer: Option<Address>,
```
colinlyguo, line 344: Would it be purer to only fetch `authorized_signer` based on the system contract's slot? But let me ask the rationale of adding this config first.

frisitano:
> LGTM. After this PR, is an L1 node (with system contract slot correctly initialized) needed for local running?

Not needed, you just have to be specific about how you configure the node. There are 2 ways you can configure the nodes such that L1 provider isn't needed:
1) Set the consensus algorithm to be `noop` (`--consensus.algorithm=noop`)
2) Set the consensus algorithm to be `system-contract` (this is default so no cli arguments requied) and then also set the authorized signer address `--consensus.authorized-signer=ADDRESS`)

tests/launch_rollup_node_follower.bash
```diff
@@ -9,4 +9,5 @@ exec rollup-node node --chain dev --datadir=/l2reth --metrics=0.0.0.0:6060 --net
   --txpool.pending-max-count=1000 \
   --builder.gaslimit=20000000 \
   --rpc.max-connections=5000 \
-  --trusted-peers enode://3983278a7cab48862d9ab3187278edf376a0736a7deb55472a5650592f6922ce626a1ea7d74b77b9a679694b343f5e93ea97d5d60a9db4e4b51bb0c23a36d01b@rollup-node-sequencer:30303
+  --trusted-peers enode://3983278a7cab48862d9ab3187278edf376a0736a7deb55472a5650592f6922ce626a1ea7d74b77b9a679694b343f5e93ea97d5d60a9db4e4b51bb0c23a36d01b@rollup-node-sequencer:30303 \
+  --consensus.algorithm=noop
```
frisitano, line 13: @colinlyguo I am setting the consensus algorithm to `noop` for this PR which is the same behaviour as we previously had from use of the `--test` flag. However, I think it would be good to migrate to `system-contract` consensus once we have anvil introduced from your opeb PR.

crates/node/src/args.rs
frisitano: This is primarily for testing purposes, such that we can run unit tests without an L1 provider. I agree with the general sentiment. How about we put this behind a `test-utils` feature flag?

crates/node/src/args.rs
```diff
@@ -202,13 +223,23 @@ impl ScrollRollupNodeConfig {
...
+                    provider
+                        .authorized_signer(node_config.address_book.system_contract_address)
+                        .await?
+                } else {
+                    return Err(eyre::eyre!(
+                        "System contract consensus requires either an authorized signer or a L1 provider URL"
+                    ));
+                };
+                Box::new(SystemContractConsensus::new(authorized_signer))
+            }
```
greged93, line 242: just an idea but should we maybe implement `fn consensus(&self)` on `ConsensusArgs`? Might be a nice way to encapsulate logic.

crates/node/src/args.rs
```diff
@@ -45,6 +45,9 @@ pub struct ScrollRollupNodeConfig {
     /// Whether the rollup node should be run in test mode.
```
greged93, line 45: should we maybe add a display of the current config used at start up? similar to how we log forks and config in Reth.

greged93 (dismissed):
two nits but feel free to merge

crates/node/src/args.rs
frisitano: I tried to do this but it turned out to introduce too much complexity. I think we should leave it as is for now and then we can refactor in the future.

crates/node/src/args.rs
frisitano: added https://github.com/scroll-tech/rollup-node/pull/231/commits/fbc8e41bf33b0db18ded614b22a2b98b022153bb

crates/node/src/args.rs
frisitano: added https://github.com/scroll-tech/rollup-node/pull/231/commits/fbc8e41bf33b0db18ded614b22a2b98b022153bb

greged93 (approved):
lgtm!

crates/node/src/args.rs
colinlyguo: Makes sense.

PR #241, opened 2025-08-07
feat: verify L1 message index

### LINKED ISSUE #236: [Indexer] Verify L1 message index
Verify that we skip L1 messages with an index that has already been indexed. In the current implementation of `DatabaseOperations::insert_l1_message(...)`, we update the L1 message on conflict. 

This should be reconsidered and we should not store L1 messages with an index that as already been seen.

varun-doshi:
Fixes #236 

Now `insert_l1_message` will not update column values on receiving l1_message with and index which is already present in the DB i.e on conflict.

Test added `test_database_duplicate_l1_index` which confirms the updated logic (this test would have failed previously)

### DIFF
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -179,7 +179,7 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
             .exec(self.get_connection())
             .await;
 
-        if let Err(DbErr::RecordNotInserted) = result {
+        if matches!(result, Err(DbErr::RecordNotInserted)) {
             tracing::error!(target: "scroll::db", queue_index = l1_index, "L1 message already exists");
             Ok(())
         } else {


crates/database/db/src/operations.rs
```diff
@@ -170,25 +170,23 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
+                    .do_nothing_on([models::l1_message::Column::QueueIndex])
                     .to_owned(),
             )
             .exec(self.get_connection())
-            .await?;
+            .await;
+
+        if let Err(err) = result {
+            tracing::info!("L1 message with queue_index {} already exists: {}", l1_index, err);
+        }
```
greged93, line 188: I'm guessing you aren't propagating the error here because you were getting a `DbErr::RecordNotInserted`? 
I don't think we should catch all errors, probably better to specifically catch the error you get if you try to insert and the record is already present, in other cases you should propagate the error.

greged93 (commented):
thanks a lot for the PR! just a comment on the error catching, otherwise looks good!

crates/database/db/src/operations.rs
varun-doshi: Ahh yes, makes sense

crates/database/db/src/operations.rs
```diff
@@ -170,25 +170,32 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
-                        models::l1_message::Column::L1BlockNumber,
-                        models::l1_message::Column::GasLimit,
-                        models::l1_message::Column::To,
-                        models::l1_message::Column::Value,
-                        models::l1_message::Column::Sender,
-                        models::l1_message::Column::Input,
-                    ])
+                    .do_nothing_on([models::l1_message::Column::QueueIndex])
                     .to_owned(),
             )
```
greged93, line 182: I think we can replace this by `on_conflict_do_nothing`, which will set a do nothing for all primary keys of the model. For the l1 message table, that is queue index.

crates/database/db/src/operations.rs
```diff
@@ -170,25 +170,32 @@ pub trait DatabaseOperations: DatabaseConnectionProvider {
...
+                DbErr::RecordNotInserted => {
+                    tracing::info!(
+                        "L1 message with queue_index {} already exists: {}",
+                        l1_index,
+                        err
+                    );
+                }
+                _ => return Err(err.into()),
+            }
+        }
```
greged93, line 197:
```suggestion
if let Err(DbErr::RecordNotInserted) = res {
    tracing::error!(target: "scroll::db", queue_index = l1_index, "L1 message already exists");
    Ok(())
} else {
    Ok(res.map(|_| ())?)
}
```

greged93 (commented):
lasts fix then we merge

Issue #141, opened 2025-05-27
[Test] Integration test cases


frisitano:
# Overview
To increase confidence in the rollup-node, we should implement the following test cases. Note that this list is not exhaustive and should be expanded on:
- [ ] Batches received from L1 should correctly update the L2 safe block head.
- [ ] Batches that have been finalized on L1 should update the L2 finalized block head.
- [ ] L1 reorg that invalidates an L1 message which has been included in a block should result in an L2 reorg.
- [ ] L1 reorg that invalidates a batch should result in the safe head reverting to the latest L2 block associated with the previous batch.
- [ ] Invalid block that is gossiped over the network should be rejected and the peer should be penalized.
- [ ] Block with invalid signature that is gossiped over the network should be rejected and the peer should be penalized.

PR #243, opened 2025-08-11
feat(e2e tests): invalid signature penalizes peer

jonastheis:
This PR addresses the following item of https://github.com/scroll-tech/rollup-node/issues/141:

> Block with invalid signature that is gossiped over the network should be rejected and the peer should be penalized.


greged93 (approved):

greged93 (approved):
lgtm!

Issue #217, opened 2025-07-24
Add support for custom genesis configuration in rollup-node


colinlyguo:
Currently, the `--chain` flag only supports predefined networks like `scroll-mainnet`.

For spinning up a reth node with customized genesis (e.g., when all forks are enabled at genesis), we need additional configuration options.

**Feature Request:**
Support launching nodes with a custom genesis file via a `--genesis` flag or similar configuration option.

PR #250, opened 2025-08-12
feat: support local genesis file

colinlyguo:
Closes: https://github.com/scroll-tech/rollup-node/issues/217.


crates/primitives/src/node/config.rs
```diff
@@ -81,14 +83,86 @@ pub struct ScrollAddressBook {
...
+        let l1_message_queue_v2_deployment_block = l1_config
+            .get("l1MessageQueueV2DeploymentBlock")
+            .and_then(|v| v.as_u64())
+            .ok_or_else(|| eyre::eyre!("Invalid or missing 'l1MessageQueueV2DeploymentBlock'"))?;
+
+        let system_contract_address = system_contract
+            .get("system_contract_address")
+            .and_then(|v| v.as_str())
+            .and_then(|s| s.parse().ok())
+            .ok_or_else(|| eyre::eyre!("Invalid or missing 'system_contract_address'"))?;
```
greged93, line 146: if you go to the `fn build` method on the `ScrollRollupNodeConfig`, you should see that we have a `ChainConfig<Config = ScrollChainConfig>` trait bound on the chain spec. If you add this here, you can skip this and just do `chain_spec.chain_config()`.

greged93 (commented):
small comment otherwise looks good!

crates/primitives/src/node/config.rs
colinlyguo: Maybe we can keep the current implementation.

yiweichi (dismissed):
lgtm!

frisitano (approved):
lgtm

yiweichi (approved):

PR #253, opened 2025-08-13
Improve L1 message reorg testing

jonastheis:
This PR mainly does the following:
- refactors `can_handle_reorgs_while_sequencing` into `can_handle_l1_message_reorg` to test not only sequencing in the presence of L1 reorg but also behavior of a follower node. Made the test much more explicit to assert certain node states
- adds a lot of new convenience functionality for e2e tests for waiting for specific events, assert node state, etc
- adjusts `default_sequencer_test_scroll_rollup_node_config` so that by default the sequencer in tests does not issue blocks automatically anymore. This is so that tests become more predictable. 
- fixes issue where an L1 reorg was not properly handled within the engine driver, leading to no L2 reorg taking place even if an L1 message was reorged. (this was discovered by refactoring the `can_handle_reorgs_while_sequencing` into `can_handle_l1_message_reorg` and making sure that node states are properly tested). 
- tests "L1 reorg that invalidates an L1 message which has been included in a block should result in an L2 reorg." #141

### DIFF
--- a/crates/engine/src/driver.rs
+++ b/crates/engine/src/driver.rs
@@ -129,12 +129,22 @@ where
         reorged_unsafe_head: Option<BlockInfo>,
         reorged_safe_head: Option<BlockInfo>,
     ) {
-        // On an unsafe head reorg: clear the payload building future, reset the unsafe head and
-        // drop the engine future if it's a `NewPayload` or `BlockImport` with block number > L2
-        // reorged number.
+        // On an unsafe head reorg.
         if let Some(l2_head_block_info) = reorged_unsafe_head {
+            // clear the payload building future.
             self.payload_building_future = None;
+
+            // retain only blocks from chain imports for which the block number <= L2 reorged
+            // number.
+            for chain_import in &mut self.chain_imports {
+                chain_import.chain.retain(|block| block.number <= l2_head_block_info.number);
+            }
+
+            // reset the unsafe head.
             self.set_head_block_info(l2_head_block_info);
+
+            // drop the engine future if it's a `NewPayload` or `BlockImport` with block number >
+            // L2 reorged number.
             if let Some(MeteredFuture { fut, .. }) = self.engine_future.as_ref() {
                 match fut {
                     EngineFuture::ChainImport(WithBlockNumber { number, .. })


greged93 (approved):
lgtm! Would like to have a review from @frisitano as well since I have implemented the L1 reorg handling in the engine driver.

crates/engine/src/driver.rs
```diff
@@ -121,6 +121,57 @@ where
...
+        // On an unsafe head reorg: clear the payload building future, reset the unsafe head and
+        // drop the engine future if it's a `NewPayload` or `BlockImport` with block number > L2
+        // reorged number.
+        if let Some(l2_head_block_info) = reorged_unsafe_head {
+            self.payload_building_future = None;
+            self.set_head_block_info(l2_head_block_info);
+            if let Some(MeteredFuture { fut, .. }) = self.engine_future.as_ref() {
+                match fut {
+                    EngineFuture::ChainImport(WithBlockNumber { number, .. })
+                        if number > &l2_head_block_info.number =>
```
frisitano, line 141: Should this be `>=`? I think it's fine as is but just want to confirm.

crates/engine/src/future/mod.rs
```diff
@@ -112,7 +112,11 @@ impl EngineFuture {
     where
         EC: ScrollEngineApi + Unpin + Send + Sync + 'static,
     {
-        Self::ChainImport(Box::pin(handle_chain_import(client, chain_import, fcs)))
+        let highest_block_number = chain_import.chain.last().unwrap().number;
+        Self::ChainImport(WithBlockNumber::new(
+            highest_block_number,
```
frisitano, line 117: Should we be using `highest_block_number ` or the `lowest_block_number`?

crates/engine/src/driver.rs
```diff
@@ -121,6 +121,57 @@ where
...
+        &mut self,
+        l1_block_number: u64,
+        reorged_unsafe_head: Option<BlockInfo>,
+        reorged_safe_head: Option<BlockInfo>,
+    ) {
+        // On an unsafe head reorg: clear the payload building future, reset the unsafe head and
+        // drop the engine future if it's a `NewPayload` or `BlockImport` with block number > L2
+        // reorged number.
+        if let Some(l2_head_block_info) = reorged_unsafe_head {
+            self.payload_building_future = None;
```
frisitano, line 136: Do we also need to filter `chain_imports` here?

frisitano (commented):
Added some questions inline.

crates/engine/src/driver.rs
greged93: Isn't `l2_head_block_info` the l2 head to which we reorged due to the L1 reorg? Meaning if we are still importing it it should be valid right?

crates/engine/src/future/mod.rs
greged93: Probably the lowest indeed, as this is relevant to L2 reorgs.

frisitano: After thinking about this, I think that the highest block number is correct.

jonastheis: Let's assume we have a chain import running from L2 block 5 to 15. And a reorg at L2 block 10 happens:
1. `5<=10`: If we take the lowest block the chain import is not cancelled
2. `15>10`: if we take the highest block the chain import is cancelled

Should be 2.

PR #256, opened 2025-08-14
add e2e test follows_most_recent_chain

jonastheis:


frisitano:
We have the following. Can you review if we need to add some more please?

https://github.com/scroll-tech/rollup-node/blob/f0c0fe701c0bd56b1814e7391cf339adef97eb9e/crates/node/tests/sync.rs#L541-L599

jonastheis:
I think one case that is missing: the block of the new canonical chain at fork height is older than the block in the new fork at the same height. however, the canonical chain is still the former (and has newer timestamps in the future).

jonastheis:
Yeah exactly. It was one of the cases that I encountered when switching between l2geth sequencer and l2reth sequencer but could happen in practice as well

frisitano:
> Yep

I'm just trying to understand how this could occur in practice?

PR #259, opened 2025-08-18
feat: cache incoming blocks to prevent duplication

### LINKED ISSUE #251: [Network] Avoid processing the same (block, signature) tuple multiple times
When configuring the node with `enable_eth_scroll_wire_bridge == true` a node _always_ receives blocks twice:
1. via `scroll wire`
2. via `eth wire`

Even if a node is only running with `scroll wire` it might receive the same (block,signature) tuple multiple times from different neighbors or a malicious neighbor might keep sending the same over and over again. 

Currently, we will process the block multiple times:
- https://github.com/scroll-tech/rollup-node/blob/4041ba70c30b92184beddc54cc77dfd49d1590b8/crates/manager/src/manager/mod.rs#L206 
- `NewBlockReceived` is fired multiple times
- signature verification is done multiple times
- block is inserted multiple times to DB, etc

This is sub-optimal and also confusing when testing (due to blocks arriving multiple times).

To solve this, we should introduce a global LRU cache in the `network` crate which tracks already seen `(block, signature)` tuples. 
- Upon receipt of a block (either via `scroll wire` or `eth wire` we add the tuple into the cache
- If we receive the same tuple again we don't process the block
- This is additional to the peer level LRU cache: https://github.com/scroll-tech/rollup-node/blob/4041ba70c30b92184beddc54cc77dfd49d1590b8/crates/scroll-wire/src/manager.rs#L28. There will be some interactions between the two caches: 
  - Currently we mark a peer as having seen a block in 2 places:
    - after successful block import: https://github.com/scroll-tech/rollup-node/blob/4041ba70c30b92184beddc54cc77dfd49d1590b8/crates/network/src/manager.rs#L175
    - after successful forwarding to peer: https://github.com/scroll-tech/rollup-node/blob/4041ba70c30b92184beddc54cc77dfd49d1590b8/crates/scroll-wire/src/manager.rs#L50
  - We need to add to the peer's cache when we don't process a block (ie step 2)
  - We might consider penalizing a peer if it sends already known information. However, it seems that even in `reth` this is not done currently: https://github.com/scroll-tech/rollup-node/blob/4041ba70c30b92184beddc54cc77dfd49d1590b8/crates/scroll-wire/src/manager.rs#L50

**Open questions:**
1. How big should the LRU cache be?

varun-doshi:
Fixes #251 

### Changes:
- Introduces a `blocks_seen: LruCache<(B256, Signature)>` in `ScrollNetworkManager` which is responsible to caching (block_hash, signature) pair
- Checks for duplicate incoming (block_hash, sig) in 2 places:
  - `on_scroll_wire_event`
  - `handle_eth_wire_block`
- If it is a duplicate block ,it will not process that further.
- If it not a duplicate block, it will add cache to two places: `blocks_seen` and per peer cache. Then will continue to process that block
- Note: currently cache size for `block_seen` is set to 100 which is same as that of the peer cache(will update this as advised)

### To-do:
- punish peers who send duplicate blocks


jonastheis:
@varun-doshi what's the status with this PR?

Issue #258, opened 2025-08-18
[Network] `eth-wire` total difficulty


frisitano:
# Overview
We must set an appropriate total difficulty when propagating blocks over the `eth-wire` protocol. The total difficulty is set here:
https://github.com/scroll-tech/rollup-node/blob/bac9dcd3bc7b36ba1449007ed06f3ff0e3eed1b4/crates/network/src/manager.rs#L121-L126
We should introduce a constant such that the total difficulty aligns with the total difficulty of l2geth and is monotonically increasing. This can be done by introducing a constant value when computing the total difficulty. 

# Constants
Mainnet: 14906960.
Sepolia: 8484488.

# Context 
since system config consensus [requires](https://github.com/scroll-tech/go-ethereum/blob/develop/consensus/system_contract/consensus.go#L150) header.difficulty to be 1, the gap should be stable.

PR #260, opened 2025-08-19
fix: total difficulty logic

frisitano:
# Overview
This PR fixes the total difficulty that is set when propagating blocks over the `eth-wire` protocol. This is achieved by adding a cosntant value to the block number such that the total difficulty aligns with that of l2geth.

closes: #258


crates/network/src/manager.rs
```diff
@@ -26,6 +27,14 @@ use tokio::sync::mpsc::{self, UnboundedReceiver};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tracing::trace;
 
+/// The constant value that must be added to the block number to get the total difficulty for Scroll
```
jonastheis, line 30: Shouldn't these constants better live in `args.rs` or wherever other config/contants are? Then we could just pass the result from `td_constant` into the network and store in `td_constant`. would keep the separation a bit cleaner as the network doesn't need the chain config

Issue #261, opened 2025-08-19
[Network] eth-wire signature


frisitano:
# Overview

## Database Model
Introduce a new table in the database, which maps `block_number -> signature`. 

There are two steps to this:
1. Define a new migration, from `crates/database` run `sea-orm-cli migrate generate block_signature`. Define the model in the migration file that is created. See example below:
https://github.com/scroll-tech/rollup-node/blob/22f5ee35ab4c566211de98a6a258c47003d2b439/crates/database/migration/src/m20220101_000001_create_batch_commit_table.rs#L1-L44
2. Implement a corresponding ORM model following the example below for the `block_signature` table.
https://github.com/scroll-tech/rollup-node/blob/22f5ee35ab4c566211de98a6a258c47003d2b439/crates/database/db/src/models/block_data.rs#L1-L37

## Database Operations
We must implement database operations to read and write block signatures. These should be implemented in `crates/database/db/src/operations.rs` following the example below:
```rust
/// The [`DatabaseOperations`] trait provides methods for interacting with the database.
#[async_trait::async_trait]
pub trait DatabaseOperations: DatabaseConnectionProvider + ScrollHardforks {
    /// Insert a block signature into the database.
    async fn insert_block_signature(
        &self,
        block_number: u64,
        signature: Signature,
    ) -> Result<(), DatabaseError> {
        tracing::trace!(target: "scroll::db", block_hash = ?block_hash, "Inserting block signature into database.");
        let block_signature_entry: models::block_signature::ActiveModel =
            (block_number, signature).into();
        Ok(models::block_signature::Entity::insert(block_signature_entry)
            .on_conflict(
                OnConflict::column(models::block_signature::Column::BlockNumber)
                    .update_column(models::block_signature::Column::Signature)
                    .to_owned(),
            )
            .exec(self.get_connection())
            .await
            .map(|_| ())?)
    }

    async fn get_block_signature(
        &self,
        block_number: u64,
        signature: Signature,
        timestamp: u64,
    ) -> Result<Option<Signature>, DatabaseError> {
        if self.is_euclid_active_at_timestamp(timestamp) {
            let data_hint = self
                .get_l2_block_data_hint(block_number)
                .await?
                .expect("Data hint must be present for L2 block")
                .extra_data
                .expect("extra data be present for L2 block");
        } else {
            Ok(models::block_signature::Entity::find()
                .filter(models::block_signature::Column::BlockNumber.eq(block_number as i64))
[...]

Issue #248, opened 2025-08-12
[Consensus] Missing header signature for serving to l2geth nodes


greged93:
l2geth expects a signature in the extra data field of the header, when requesting a header from p2p and also when receiving a header for announced blocks. This extra data field is removed from the header by the RN when sending it to the EN, meaning the EN cannot provide the signature on header requests.

Proposed solution: Shared DB, RN writes signature at gossip from eth-wire and scroll-wire, EN writes signature when processing GetBlockHeaders (in the HeaderDownloaded?). EN reads when it needs the signature for serving `GetBlockHeaders` (in the `EthRequestHandler`).

PR #267, opened 2025-08-24
fix: eth66 serve signature

yiweichi:
This PR fix the issue reth doesn't include sig during block request, l2geth will fail the block validation of blocks requested from reth.
The solution is that we persist block signature to the database of rollup-node, retrieve and serve it during handling eth66 block request.

Corresponding issue: https://github.com/scroll-tech/rollup-node/issues/261
Corresponding reth PR: https://github.com/scroll-tech/reth/pull/331

Closes: #261 
Closes: #248


crates/manager/src/manager/mod.rs
```diff
@@ -526,6 +539,18 @@ where
...
+                    let block_hash = block.hash_slow();
+                    tokio::task::block_in_place(|| {
+                        tokio::runtime::Handle::current().block_on(async move {
+                            if let Err(err) = database.insert_signature(block_hash, signature).await {
+                                error!(target: "scroll::node::manager", ?err, "Failed to store block signature");
+                            }
+                        })
+                    });
+
                     this.network.handle().announce_block(block, signature);
```
jonastheis, line 554: We could do the above in `ScrollNetworkManager:announce_block` then there's no need to duplicate it. 

We call `announce_block` for both, sequenced and successfully processed blocks

crates/database/db/src/models/block_signature.rs
```diff
@@ -0,0 +1,47 @@
+use alloy_primitives::{Signature, B256};
+use sea_orm::{entity::prelude::*, ActiveValue};
+
+/// A database model that represents a block signature.
```
jonastheis, line 4: Maybe add to the comment here that it can also be deprecated once l2geth is deprecated and for what purpose we're storing it

jonastheis (commented):
I think the only thing missing is this: https://github.com/scroll-tech/rollup-node/issues/261#issuecomment-3204828703

frisitano (approved):
Looks good. Let's get the reth PR merged first and then repoint deps.

crates/manager/src/manager/mod.rs
yiweichi: Yes, that's my original thought on this.

crates/manager/src/manager/mod.rs
```diff
@@ -225,7 +228,19 @@ where
...
-            self.engine.handle_block_import(block_with_peer);
+            self.engine.handle_block_import(block_with_peer.clone());
+
+            // TODO: remove this once we deprecate l2geth.
+            // Store the block signature in the database
+            tokio::task::block_in_place(|| {
+                tokio::runtime::Handle::current().block_on(async move {
+                    let block_hash = block_with_peer.block.hash_slow();
+                    let signature = block_with_peer.signature;
+                    self.database.insert_signature(block_hash, signature).await
```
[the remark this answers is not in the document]
yiweichi: what if the insert_signature operation failed? and we will have block on db without signature, l2geth node will drop us. would that be a problem?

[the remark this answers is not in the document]
yiweichi: Raised issue to track this https://github.com/scroll-tech/rollup-node/issues/277

crates/database/db/src/models/block_signature.rs
yiweichi: added this in commit [c707fe3](https://github.com/scroll-tech/rollup-node/pull/267/commits/c707fe30a6fa81d222a20882c2503b195e1e379d)

crates/signer/src/signature.rs
```diff
@@ -0,0 +1,265 @@
+//! Custom signature implementation that wraps alloy_primitives::Signature
```
frisitano, line 1: Just wondering if it would be cleaner to implement a trait on `Signature` instead of wrapping it. Something like:
```rust
pub trait SignatureAsBytes {
    pub fn sig_as_bytes(&self) -> [u8; 65];
}

impl SignatureAsBytes for alloy_primitives::Signature {

    /// Custom implementation of as_bytes 
    /// We remove the `+ 27` offset from the v value
    /// and use the original v value directly.
    #[inline]
    pub fn sig_as_bytes(&self) -> [u8; 65] {
        let mut sig = [0u8; 65];
        sig[..32].copy_from_slice(&self.inner.r().to_be_bytes::<32>());
        sig[32..64].copy_from_slice(&self.inner.s().to_be_bytes::<32>());
        sig[64] = self.inner.v() as u8;
        sig
    }
}
```

frisitano (commented):
Looks good. Left some comments inline about how we handle the signature. I think it would be cleaner and simpler to introduce and implement a trait on `alloy_primitives::Signature` what do you think?

crates/signer/src/signature.rs
yiweichi: Good idea!

yiweichi: Done

frisitano (commented):
Added some initial comments inline.

crates/node/src/builder/network.rs
```diff
@@ -0,0 +1,347 @@
...
+        let chain_spec = ctx.chain_spec();
+        let authorized_signer = if self.signer.is_none() {
+            match chain_spec.chain().named() {
+                Some(NamedChain::Scroll) => Some(SCROLL_MAINNET_SIGNER),
+                Some(NamedChain::ScrollSepolia) => Some(SCROLL_SEPOLIA_SIGNER),
+                _ => None,
+            }
+        } else {
+            self.signer
+        };
```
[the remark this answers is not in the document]
yiweichi: Make sense, done in [607032f](https://github.com/scroll-tech/rollup-node/pull/267/commits/607032f4e63d892f62218bea40937e87346b5a05)

crates/database/db/src/db.rs
```diff
@@ -25,7 +29,7 @@ const ACQUIRE_TIMEOUT_SECS: u64 = 5;
 /// operations defined in [`crate::DatabaseOperations`]. Atomic operations can be performed using
 /// the [`Database::tx`] method which returns a [`DatabaseTransaction`] that also implements the
 /// [`DatabaseConnectionProvider`] trait and also the [`crate::DatabaseOperations`] trait.
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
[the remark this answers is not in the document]
yiweichi: let me remove it.

crates/database/db/src/db.rs
```diff
@@ -13,7 +14,10 @@ use sea_orm::{
 const BUSY_TIMEOUT_SECS: u64 = 5;
 
 /// The maximum number of connections in the database connection pool.
-const MAX_CONNECTIONS: u32 = 10;
+const MAX_CONNECTIONS: u32 = 500;
```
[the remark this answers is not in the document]
yiweichi: I think I increase this because it used to always failed to get db connection when inserting the block signature. I remember after increased this value to 250, the issue eventually go away. So I set the value as 500 eventually out of redundancy.

.github/workflows/test.yaml
```diff
@@ -32,9 +33,12 @@ jobs:
           tool: nextest@0.9.98
       - name: Run unit tests
         run: cargo nextest run --all-features --workspace --locked -E '!kind(test)'
+        env:
+          CARGO_BUILD_JOBS: 4
```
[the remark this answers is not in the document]
yiweichi: That was for testing, I would have thought the CI fails because of this. Reverted this.

frisitano (approved):
lgtm

PR #271, opened 2025-08-25
add logic to skip reorg on l1 consolidation when l1 is not yet synced

frisitano:


frisitano:
closing in favour of an alternative solution.

Issue #273, opened 2025-08-26
[Watcher] Process batches differently based on sync or live mode.


greged93:
Add 2 confirmation modes, one for backfill/initial sync, another for live sync. 
- Initial sync: Only process batches once they’re really finalized (i.e. FinalizeBatch event is finalized on L1). 
- Live sync: Process committed batches, reorg on reverts.

PR #275, opened 2025-08-26
fix: handle batch revert

greged93:
In place of handling batch reverts, we only pass L1 finalized finalized batches to the derivation pipeline. This is a temporary solution and should be replaced in time by #273.

I ran this PR on Sepolia, consolidating up to block 100k and didn't encounter the previous issue with reverted batches.

### DIFF
--- a/crates/primitives/src/block.rs
+++ b/crates/primitives/src/block.rs
@@ -108,8 +108,8 @@ impl<T: Future + Unpin> Future for WithBlockNumber<T> {
 /// A type alias for a wrapper around a type to which a finalized batch information is attached.
 pub type WithFinalizedBatchInfo<T> = WithBatchInfo<T>;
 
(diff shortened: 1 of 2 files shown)


crates/chain-orchestrator/src/lib.rs
```diff
@@ -699,23 +711,30 @@ impl<
...
         database.finalize_batch(batch_hash, block_number).await?;
 
-        // check if the block where the batch was finalized is finalized on L1.
         let mut finalized_block = None;
+        let mut finalized_batch = None;
+
+        // check if the block where the batch was finalized is finalized on L1.
         let l1_block_number_value = l1_block_number.load(Ordering::Relaxed);
-        if l1_block_number_value > block_number {
+        if l1_block_number_value >= block_number {
```
greged93, line 727: I think this was a mistake.

frisitano (approved):

PR #278, opened 2025-08-27
feat: introduce watcher back pressure in the rollup node manager

frisitano:
# Overview
This PR introduces back pressure in the rollup node manager such that we only progress with consuming additional L1 messages when the chain orchestrator and engine driver have a reasonable number of pending futures in their queues.


greged93 (approved):
lgtm! We'll need to find the right balance for the chain orchestrator and engine, weighing memory usage against throughput improvements, but we can do this as we go.

PR #281, opened 2025-08-28
feat: add indexes and improve SQL default config

greged93:


frisitano (changes requested):
Added some comments inline.

crates/database/db/src/db.rs
```diff
@@ -22,18 +34,20 @@ pub struct Database {
...
-        Ok(Self { connection })
+        let options = SqliteConnectOptions::from_str(database_url)?
+            .create_if_missing(true)
+            .journal_mode(sea_orm::sqlx::sqlite::SqliteJournalMode::Wal)
+            .busy_timeout(Duration::from_secs(BUSY_TIMEOUT_SECS))
+            .foreign_keys(true)
+            .synchronous(sea_orm::sqlx::sqlite::SqliteSynchronous::Normal);
+
+        let sqlx_pool = sea_orm::sqlx::sqlite::SqlitePoolOptions::new()
+            .max_connections(MAX_CONNECTIONS)
```
greged93, line 45: not very important but I believe the default for max connections is already 10.

greged93 (commented):
just a small comment otherwise ltgm! (can't approve because I'm the PR author).

PR #287, opened 2025-08-29
feat: skip empty blocks

### LINKED ISSUE #282: [Sequencer] Do not build empty blocks
# Overview
We should implement logic that skips the production of empty blocks. This should be done to avoid the unnecessary overhead of proving and DA of empty blocks.

Potential soltuions:

1. Build blocks as is normally, done and then skip if the block has no transactions
2. Have handles to L1 messages and transaction pool, and do not build a block if there are no transactions.

The consensus is that option 1 is the preferred and simpler solution.

varun-doshi:
Fixes #282

### DIFF
--- a/crates/node/src/args.rs
+++ b/crates/node/src/args.rs
@@ -431,19 +431,14 @@ impl ConsensusArgs {
 }
 
 /// The consensus algorithm to use.
-#[derive(Debug, clap::ValueEnum, Clone, PartialEq, Eq)]
+#[derive(Debug, Default, clap::ValueEnum, Clone, PartialEq, Eq)]
 pub enum ConsensusAlgorithm {
     /// System contract consensus with an optional authorized signer. If the authorized signer is
     /// not provided the system will use the L1 provider to query the authorized signer from L1.
+    #[default]
     SystemContract,
     /// No-op consensus that does not validate blocks.
     Noop,
-}
-
-impl Default for ConsensusAlgorithm {
-    fn default() -> Self {
-        Self::SystemContract
-    }
 }
 
 /// The engine driver args.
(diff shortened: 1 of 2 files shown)


jonastheis (commented):
Looks good but would be better to have the explicit option via command line flag to enable empty block production: https://github.com/scroll-tech/rollup-node/issues/282#issuecomment-3240612621

varun-doshi:
> Looks good but would be better to have the explicit option via command line flag to enable empty block production: [#282 (comment)](https://github.com/scroll-tech/rollup-node/issues/282#issuecomment-3240612621)

Understood, will do

crates/node/src/args.rs
```diff
@@ -78,6 +78,9 @@ pub struct ScrollRollupNodeConfig {
     /// The gas price oracle args
     #[command(flatten)]
     pub gas_price_oracle_args: GasPriceOracleArgs,
+    /// Enable empty blocks
+    #[arg(long)]
+    pub allow_empty_blocks: bool,
```
frisitano, line 83: This will default to false (if omitted from node startup command) - is that desired behaviour @jonastheis?

frisitano (commented):
Looks good, thanks for this! I've left one comment inline which I would like clarifcation from @jonastheis on.

As a follow up I think it would be good to introduce an RPC extension that allows us to toggle this setting following the pattern in #289.

jonastheis (commented):
This [CI failure](https://github.com/scroll-tech/rollup-node/actions/runs/17427236128/job/49598248667#step:9:87) might be a real one: for the docker integration tests we probably need to run the node now with the flag to enable empty blocks specifically.

jonastheis (approved):
lgtm!

jonastheis:
@varun-doshi best to always run `make pr` before committing to make sure that all tests and lints pass

PR #288, opened 2025-08-29
feat: finalize all batches up to received index

greged93:
Fixes the incorrect logic for handling batch finalization logs. Also adds logging for the scroll watcher to monitor the log filtering.


frisitano (approved):
lgtm

PR #291, opened 2025-09-01
feat: update batch processing logic

frisitano:
# Overview
This PR modifies the way in which batches are triggered to be sent to the derivation pipeline. 

# Changes
1. Introduction of a `processed` column on the `batch_commit` table and associated migrations.
2. Modify startup logic to set the `start_l1_block_number` using the correct `block_number` from the last fully processed batch.
3. Modify logic for selecting and pushing batches into the derivation pipeline.

### DIFF
--- a/crates/chain-orchestrator/src/event.rs
+++ b/crates/chain-orchestrator/src/event.rs
@@ -43,8 +43,8 @@ pub enum ChainOrchestratorEvent {
     /// A batch has been finalized returning an optional finalized L2 block. Also returns a
     /// [`BatchInfo`] if the finalized event occurred in a finalized L1 block.
     BatchFinalized(Option<WithFinalizedBlockNumber<BatchInfo>>, Option<BlockInfo>),
-    /// An L1 block has been finalized returning the L1 block number, the list of finalized batches
-    /// and an optional finalized L2 block.
+    /// An L1 block has been finalized returning the L1 block number and the list of finalized
+    /// batches.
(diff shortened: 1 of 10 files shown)


crates/chain-orchestrator/src/event.rs
```diff
@@ -45,7 +45,7 @@ pub enum ChainOrchestratorEvent {
     BatchFinalized(Option<WithFinalizedBlockNumber<BatchInfo>>, Option<BlockInfo>),
     /// An L1 block has been finalized returning the L1 block number, the list of finalized batches
     /// and an optional finalized L2 block.
```
jonastheis, line 47: adjust comment

crates/chain-orchestrator/src/lib.rs
```diff
@@ -615,35 +607,18 @@ impl<
...
-
-        // get the finalized block for the batch.
-        let finalized_block = if let Some(info) = finalized_batches.last() {
-            Self::fetch_highest_finalized_block(database, info.hash, l2_block_number).await?
-        } else {
-            None
-        };
+        // Get all unprocessed batches that have been finalized by this L1 block finalization.
+        let finalized_batches =
+            database.fetch_and_update_unprocessed_finalized_batches(block_number).await?;
```
jonastheis, line 616: We mark the batches as processed before they are actually handled by the derivation pipeline. How do we know which batch was processed by the derivation pipeline in case the node crashes?

jonastheis (approved):

PR #295, opened 2025-09-02
fix: l1 message indexing

frisitano:
# Overview
This PR modifies the indexing of L1 messages such that we correctly handle the boundary from V1 to V2. Additionally, we update the derivation pipeline to leverage the `l1_v2_message_queue_start_index` when it encounters a batch with a `prev_l1_message_queue_hash == B256::ZERO` (the first batch after the migration to the V2 message queue).

# Testing
I have tested this on Sepolia and it is working fine. I would propose we open an issue to improve test coverage of this, but I would like to proceed with implementing an RPC endpoint for rolling back the database to an L1 block, as it is more pressing.

### DIFF
--- a/crates/chain-orchestrator/src/lib.rs
+++ b/crates/chain-orchestrator/src/lib.rs
@@ -686,10 +686,10 @@ impl<
 }
 
 /// Computes the queue hash by taking the previous queue hash and performing a 2-to-1 hash with the
-/// current transaction hash using keccak. It then applies a mask which is required for the kzg
-/// field used for blobs on ethereum. For the first message in the queue, the previous queue
-/// hash is zero. If the L1 message queue index is before migration to `L1MessageQueueV2`,
-/// the queue hash will be None.
+/// current transaction hash using keccak. It then applies a mask to the last 32 bits as these bits
+/// are used to store the timestamp at which the message was enqueued in the contract. For the first
+/// message in the queue, the previous queue hash is zero. If the L1 message queue index is before
+/// migration to `L1MessageQueueV2`, the queue hash will be None.
 ///
 /// The solidity contract (`L1MessageQueueV2.sol`) implementation is defined here: <https://github.com/scroll-tech/scroll-contracts/blob/67c1bde19c1d3462abf8c175916a2bb3c89530e4/src/L1/rollup/L1MessageQueueV2.sol#L379-L403>
 async fn compute_l1_message_queue_hash(


crates/chain-orchestrator/src/lib.rs
```diff
@@ -534,15 +538,14 @@ impl<
...
+            L1Notification::L1Message { message, block_number, block_timestamp: _ } => {
                 ChainOrchestratorFuture::HandleL1Message(self.handle_metered(
                     ChainOrchestratorItem::L1Message,
                     Box::pin(Self::handle_l1_message(
                         self.database.clone(),
-                        self.chain_spec.clone(),
                         message,
                         block_number,
-                        block_timestamp,
+                        self.l1_v2_message_queue_start_index,
```
jonastheis, line 548: Let's pass it earlier as it is a non-changing parameter?

crates/chain-orchestrator/src/lib.rs
```diff
@@ -624,26 +627,25 @@ impl<
...
 
-        let queue_hash = if chain_spec
-            .scroll_fork_activation(ScrollHardfork::EuclidV2)
-            .active_at_timestamp_or_number(block_timestamp, l1_block_number) &&
-            l1_message.queue_index > 0
-        {
+        let queue_hash = if l1_message.queue_index == l1_v2_message_queue_start_index {
+            let mut input = B256::default().to_vec();
+            input.append(&mut l1_message.tx_hash().to_vec());
+            Some(keccak256(input) & L1_MESSAGE_QUEUE_HASH_MASK)
```
jonastheis, line 639: Would be good to put this in a function to avoid duplicate code and reference the contracts implementation of it.

crates/derivation-pipeline/src/lib.rs
```diff
@@ -260,10 +270,15 @@ pub async fn derive<L1P: L1Provider + Sync + Send, L2P: BlockDataProvider + Sync
     if let Some(index) = data.queue_index_start() {
         l1_provider.set_queue_index_cursor(index);
     } else if let Some(hash) = data.prev_l1_message_queue_hash() {
-        l1_provider.set_hash_cursor(*hash).await;
-        // we skip the first l1 message, as we are interested in the one starting after
-        // prev_l1_message_queue_hash.
-        let _ = l1_provider.next_l1_message().await.map_err(Into::into)?;
+        // If the message queue hash is zero then we should use the V2 L1 message queue start index.
+        if hash == &B256::ZERO {
```
jonastheis, line 274: This works bc otherwise it would be None? If so, please add a comment of this here.

jonastheis (changes requested):

crates/chain-orchestrator/src/lib.rs
jonastheis: I mean the order of function arguments

crates/chain-orchestrator/src/lib.rs
frisitano: No, we don't compute the hashes in the derivation pipeline. We just use the `prev_l1_message_queue_hash` to do a lookup to determine the start index of L1 messages to iterate from when processing a batch. In any case, I think it's cleaner to extract this to a function.

crates/derivation-pipeline/src/lib.rs
frisitano: correct, there is no queue hash of `B256::ZERO` indexed in the database.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -702,6 +685,39 @@ impl<
     }
 }
 
+/// Computes the queue hash by taking the previous queue hash and performing a 2-to-1 hash with the
+/// current transaction hash using keccak. It then applies a mask which is required for the kzg
```
jonastheis, line 689: nit: the mask is not required for kzg. We mask it because in the contract we also store the timestamp when the message was appended to the queue in the last few bits

jonastheis (approved):

Issue #262, opened 2025-08-20
[Provider]: support config multiple blob providers


yiweichi:
Support config multiple blob providers, and round robin provides as same as l2geth. This will increase stability.

PR #305, opened 2025-09-08
feat:  enable set multiple blob provider

yiweichi:
This PR enable RN to set multiple blob providers, and round robin provides. This will increase stability. 

There is a priority order: 
1. beacon node by default (if configurated)
2. if beacon node fails, we round robin other providers

Corresponding issue: https://github.com/scroll-tech/rollup-node/issues/262


crates/node/src/args.rs
```diff
@@ -546,17 +549,18 @@ pub struct L1ProviderArgs {
 /// The arguments for the Beacon provider.
 #[derive(Debug, Default, Clone, clap::Args)]
 pub struct BeaconProviderArgs {
```
frisitano, line 551: Should we change the name of these args to 
```suggestion
pub struct BlobProviderArgs {
```

crates/node/src/args.rs
```diff
@@ -546,17 +549,18 @@ pub struct L1ProviderArgs {
...
+    pub beacon_node_url: Option<reqwest::Url>,
+    /// The URL for the s3 blob provider.
+    #[arg(long = "beacon.s3_url", id = "beacon_s3_url", value_name = "BEACON_S3_URL")]
+    pub s3_url: Option<reqwest::Url>,
+    /// The URL for the anvil blob provider.
+    #[arg(long = "beacon.anvil_url", id = "anvil_url", value_name = "ANVIL_URL")]
+    pub anvil_url: Option<reqwest::Url>,
+    /// Enable the mock blob source.
+    #[arg(long = "beacon.mock", default_value_t = false)]
+    pub mock: bool,
```
frisitano, line 563: if we rename to `BlobProviderArgs` then the `long` variants should be come `long = blob.*`

crates/node/src/args.rs
```diff
@@ -546,17 +549,18 @@ pub struct L1ProviderArgs {
...
+    #[arg(long = "beacon.beacon_node_url", id = "beacon_node_url", value_name = "BEACON_NODE_URL")]
+    pub beacon_node_url: Option<reqwest::Url>,
+    /// The URL for the s3 blob provider.
+    #[arg(long = "beacon.s3_url", id = "beacon_s3_url", value_name = "BEACON_S3_URL")]
+    pub s3_url: Option<reqwest::Url>,
+    /// The URL for the anvil blob provider.
+    #[arg(long = "beacon.anvil_url", id = "anvil_url", value_name = "ANVIL_URL")]
+    pub anvil_url: Option<reqwest::Url>,
+    /// Enable the mock blob source.
+    #[arg(long = "beacon.mock", default_value_t = false)]
```
frisitano, line 562: The default value of a Boolean is false so there's no need to add `default_value_t = false` but feel free to keep it if you think it makes the code clearer.

crates/node/src/args.rs
```diff
@@ -546,17 +549,18 @@ pub struct L1ProviderArgs {
...
-    #[arg(
-        long = "beacon.blob-source",
-        id = "beacon_blob_source",
-        value_name = "BEACON_BLOB_SOURCE",
-        default_value = "mock"
-    )]
-    pub blob_source: BlobSource,
+    /// The URL for the beacon node blob provider.
+    #[arg(long = "beacon.beacon_node_url", id = "beacon_node_url", value_name = "BEACON_NODE_URL")]
+    pub beacon_node_url: Option<reqwest::Url>,
```
frisitano, line 554: I think this is fine for now but it might be more flexible if we could parse a `Option<Vec<reqwest::Url>>`. This could support the use case where we have multiple `beacon_node_url`, i.e. self-hosted beacon node, infura, alchemy etc.

crates/providers/src/l1/blob/mod.rs
```diff
@@ -30,38 +34,131 @@ pub trait BlobProvider: Sync + Send {
...
+    pub beacon: Option<reqwest::Url>,
     /// AWS S3 blob source.
-    S3,
+    pub s3: Option<reqwest::Url>,
     /// Anvil sequencer blob source.
-    Anvil,
+    pub anvil: Option<reqwest::Url>,
+    /// Mocked source.
+    pub mock: bool,
 }
```
frisitano, line 48: I think this is fine for now but in the future if we wanted to provide multiple instances of the different provider types we may want to transition to something like this:

```rust
pub enum BlobSource {
    Beacon(reqwest::Url),
    CloudStorage(reqwest::Url),
    Anvil(reqwest::Url),
    Mock,
}

#[derive(Default)]
pub struct BlobProvidersBuilder {
    sources: Vec<BlobSource>,
}

impl BlobProvidersBuilder {
    pub fn with_blob_source(mut self, source: BlobSource) -> Self {
        self.sources.push(source);
        self
    }

    pub fn with_beacon_source(mut self, url: reqwest::Url) -> Self {
        self.sources.push(BlobSource::Beacon(url));
        self
    }

    pub fn with_cloud_source(mut self, url: reqwest::Url) -> Self {
        self.sources.push(BlobSource::CloudStorage(url));
        self
    }

    ...
}
```

frisitano (commented):
Thanks, looks good! Left some minor nits inline.

crates/node/src/args.rs
yiweichi: Good idea. Done in commit [6fd967e](https://github.com/scroll-tech/rollup-node/pull/305/commits/6fd967e5156006e981da2476c0724c04fdb7f0d9)

crates/providers/src/l1/blob/mod.rs
yiweichi: Yeah, make sense. but I’d prefer to do it then if we really had new types of provider.

crates/providers/src/l1/blob/mod.rs
```diff
@@ -30,38 +34,159 @@ pub trait BlobProvider: Sync + Send {
...
-                url.ok_or_eyre("missing url for s3 blob provider")?,
-            )),
-            Self::Anvil => Arc::new(AnvilBlobProvider::new_http(
-                url.ok_or_eyre("missing url for anvil blob provider")?,
-            )),
-        })
+impl BlobProvidersBuilder {
+    /// Returns an [`BlobProviders`].
+    pub async fn build(&self) -> eyre::Result<BlobProviders> {
+        if self.mock {
```
jonastheis, line 53: Maybe would be good to log here with an `info` or `warn` that it's running with a mock and all the other parameters are ignored

crates/providers/src/l1/blob/mod.rs
```diff
@@ -30,38 +34,159 @@ pub trait BlobProvider: Sync + Send {
...
+
+            for i in 0..self.beacon_providers.len() {
+                let provider_index = (start_index + i) % self.beacon_providers.len();
+                let provider = &self.beacon_providers[provider_index];
+
+                match provider.blob(block_timestamp, hash).await {
+                    Ok(blob) => {
+                        // Update counter to start from next provider next time if we found the blob
+                        if blob.is_some() {
+                            self.beacon_counter.store(provider_index + 1, Ordering::Relaxed);
```
jonastheis, line 147: Why only update the counter when we successfully retrieved the blob? Wouldn't that mean we always try a failing beacon node again first?

crates/providers/src/l1/blob/mod.rs
```diff
@@ -30,38 +34,159 @@ pub trait BlobProvider: Sync + Send {
...
+                        debug!(target: "scroll::providers", ?hash, ?block_timestamp, ?provider_index, ?err, "beacon provider failed to fetch blob");
+                    }
+                }
+            }
+
+            debug!(target: "scroll::providers", ?hash, ?block_timestamp, "All beacon providers failed, trying backup providers");
+        }
+
+        // Try backup providers in round-robin order
+        if self.backup_providers.is_empty() {
```
jonastheis, line 161: Why duplicate the error message here. better to invert the if imo

jonastheis (approved):

PR #312, opened 2025-09-11
fix: eth66 serve/persist signature

yiweichi:
This PR 
1. switches to using the same function as the consensus validation to recover the signer when persisting signatures. 
2. uses `sig_encode_hash` to calculate the right header hash when handle block request and block receiving.


jonastheis (dismissed):

frisitano (approved):
lgtm

Issue #319, opened 2025-09-15
[Chain Orchestrator] Retry on failure


frisitano:
# Overview
If we have a failure when processing a future in the chain orchestrator and the error is associated with some transient error, such as `DatabaseError`, `NetworkRequestError` or `RpcError`, then we should retry the future until completion. This prevents transient errors from impacting the chain orchestrator. We can do this by marking the error types as retryable or not e.g.
```rust
impl ChainOrchestratorError {
    fn can_retry(&self) -> bool {
        matches!(
            self,
            ChainOrchestratorError::DatabaseError(_) |
                ChainOrchestratorError::NetworkRequestError(_) |
                ChainOrchestratorError::RpcError(_)
        )
    }
}
```

We would then implement the retry logic in the `Stream` implementation of the `ChainOrchestrator`:
https://github.com/scroll-tech/rollup-node/blob/3d400f7820809a27c57525ef2b37638d5e00edbf/crates/chain-orchestrator/src/lib.rs#L775-L804

frisitano:
I'm thinking we can implement something like this using the tower stack. This isn't compiling at the moment due to using references for the transaction. We may need to pass an owned `TX` / `TXMut` instead. Let me know what you think.
```rust
pub enum DatabaseQuery<T> {
    Read(
        Box<
            dyn FnOnce(&TX) -> Pin<Box<dyn Future<Output = Result<T, DatabaseError>> + Send>>
                + Send,
        >,
    ),
    Write(
        Box<
            dyn FnOnce(&mut TXMut) -> Pin<Box<dyn Future<Output = Result<T, DatabaseError>> + Send>>
                + Send,
        >,
    ),
}

impl<T> DatabaseQuery<T>
where
    T: Send + 'static,
{
    pub fn read<F, Fut>(f: F) -> Self
    where
        F: FnOnce(&TX) -> Fut + Send + 'static,
        Fut: Future<Output = Result<T, DatabaseError>> + Send + 'static,
    {
        DatabaseQuery::Read(Box::new(move |tx| Box::pin(f(tx))))
    }

    pub fn write<F, Fut>(f: F) -> Self
    where
        F: FnOnce(&mut TXMut) -> Fut + Send + 'static,
        Fut: Future<Output = Result<T, DatabaseError>> + Send + 'static,
    {
        DatabaseQuery::Write(Box::new(move |tx| Box::pin(f(tx))))
    }
}

[...]

PR #322, opened 2025-09-16
feat(chain-orchestractor): support retry on failure

yiweichi:
This PR supports retry on part of chain orchestractor failures.

Currently including failure types:
- DatabaseError
- NetworkRequestError
- RpcError

Corresponding issue: https://github.com/scroll-tech/rollup-node/issues/319

### DIFF
--- a/crates/chain-orchestrator/src/lib.rs
+++ b/crates/chain-orchestrator/src/lib.rs
@@ -711,7 +711,7 @@ impl<
         let l1_message = L1MessageEnvelope::new(l1_message, l1_block_number, None, queue_hash);
 
         // Perform a consistency check to ensure the previous L1 message exists in the database.
-        let _ = Retry::default()
+        Retry::default()
             .retry("handle_l1_message", || async {
                 let tx = database.tx_mut().await?;
                 if l1_message.transaction.queue_index > 0 &&
@@ -728,7 +728,7 @@ impl<
                 tx.commit().await?;
                 Ok::<_, ChainOrchestratorError>(())
             })
-            .await;
+            .await?;
         Ok(Some(event))
     }
 
--- a/crates/chain-orchestrator/src/error.rs
+++ b/crates/chain-orchestrator/src/error.rs
@@ -3,8 +3,6 @@ use alloy_primitives::B256;
 use alloy_transport::TransportErrorKind;
 use scroll_db::{DatabaseError, L1MessageStart};
 
-use crate::retry::CanRetry;
-
 /// A type that represents an error that occurred in the chain orchestrator.
 #[derive(Debug, thiserror::Error)]
 pub enum ChainOrchestratorError {
@@ -53,24 +51,4 @@ pub enum ChainOrchestratorError {
     /// An error occurred while making a JSON-RPC request to the Execution Node (EN).
     #[error("An error occurred while making a JSON-RPC request to the EN: {0}")]
     RpcError(#[from] RpcError<TransportErrorKind>),
-}
-
-// Implement the local CanRetry trait for the external DatabaseError type so we can delegate.
-impl CanRetry for DatabaseError {
-    fn can_retry(&self) -> bool {
-        matches!(self, Self::DatabaseError(_) | Self::SqlxError(_))
-    }
-}
-
-impl CanRetry for ChainOrchestratorError {
-    fn can_retry(&self) -> bool {
-        match self {
-            // Delegate to DatabaseError's classification
-            Self::DatabaseError(db) => db.can_retry(),
-            // Network and RPC errors are generally transient
-            Self::NetworkRequestError(_) | Self::RpcError(_) => true,
-            // All others are logical/state errors: do not retry
-            _ => false,
-        }
-    }
 }
--- a/crates/chain-orchestrator/src/retry.rs
+++ b/crates/chain-orchestrator/src/retry.rs
@@ -19,11 +19,6 @@ impl Default for Retry {
     }
 }
 
-/// A trait for errors that can indicate whether an operation can be retried.
-pub trait CanRetry {
-    fn can_retry(&self) -> bool;
-}
-
 impl Retry {
     /// Creates a new [`Retry`] with the specified parameters.
     pub const fn new(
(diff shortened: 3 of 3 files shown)


crates/chain-orchestrator/src/lib.rs
```diff
@@ -48,8 +48,36 @@ pub use metrics::{ChainOrchestratorItem, ChainOrchestratorMetrics};
 const L1_MESSAGE_QUEUE_HASH_MASK: B256 =
     b256!("ffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000");
 
+/// Delay between retry attempts in milliseconds
+const RETRY_DELAY_MS: u64 = 500;
```
frisitano, line 52: This is too long to wait for a retry. If we put it into perspective, the current block time is 500 milliseconds, but we have an ambition to reduce it further. The current delay could cause lag when following the chain tip. I suggest we reduce the retry delay to 10ms.

frisitano (commented):
It looks like you have not implemented the retry logic on some critical components. Specifically:
- persist_l1_consolidated_blocks
- consolidate_validated_l2_blocks
- handle_l1_reorg

I also have some concerns about this pattern. Lets say for example that we implement the retry mechanic on:

```rust
    /// Handles a reorganization event by deleting all indexed data which is greater than the
    /// provided block number.
    async fn handle_l1_reorg(
        database: Arc<Database>,
        chain_spec: Arc<ChainSpec>,
        l1_block_number: u64,
        l2_client: Arc<P>,
        current_chain: Arc<Mutex<Chain>>,
    ) -> Result<Option<ChainOrchestratorEvent>, ChainOrchestratorError> {
        let txn = database.tx().await?;
        let UnwindResult { l1_block_number, queue_index, l2_head_block_number, l2_safe_block_info } =
            txn.unwind(chain_spec.genesis_hash(), l1_block_number).await?;
        txn.commit().await?;
        let l2_head_block_info = if let Some(block_number) = l2_head_block_number {
            // Fetch the block hash of the new L2 head block.
            let block_hash = l2_client
                .get_block_by_number(block_number.into())
                .await?
                .expect("L2 head block must exist")
                .header
                .hash_slow();
            // Remove all blocks in the in-memory chain that are greater than the new L2 head block.
            let mut current_chain_headers = current_chain.lock().await;
            current_chain_headers.inner_mut().retain(|h| h.number <= block_number);
            Some(BlockInfo { number: block_number, hash: block_hash })
        } else {
            None
        };
        Ok(Some(ChainOrchestratorEvent::L1Reorg {
            l1_block_number,
            queue_index,
            l2_head_block_info,
            l2_safe_block_info,
        }))
    }
```

Let's say we commit the database transaction but then we fail on `get_block_by_number` and attempt to retry. On the second retry the response from `txn.unwind` will be different and as such the second retry will yield an incorrect result. 

We have two options to address this:
- Hold the database transaction open until we have the final result ready. This is bad practice as we shouldn't hold a database lock whilst performing other IO operations.
- Only do read operations at the start of the function and then finaly commit all changes to the database at the end of the function when we have computed the result.

I think we should migrate to pattern 2. 


Let me know your thoughts.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -48,8 +48,36 @@ pub use metrics::{ChainOrchestratorItem, ChainOrchestratorMetrics};
...
+                    attempt = attempts,
+                    "Retrying operation due to transient error"
+                );
+                tokio::time::sleep(tokio::time::Duration::from_millis(RETRY_DELAY_MS)).await;
+            }
+            Err(e) => return Err(e),
+        }
+    }
+}
+
```
greged93, line 80: I don't know if `retry_with_backoff` is the correct strategy, as I think it could lead to issues due to the fact that the whole task is retried.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -632,63 +663,83 @@ impl<
...
-        } else {
-            None
-        };
+                // handle the case of a batch revert.
+                let new_safe_head = if affected > 0 {
+                    txn.delete_l2_blocks_gt_batch_index(prev_batch_index).await?;
+                    txn.get_highest_block_for_batch_index(prev_batch_index).await?
+                } else {
+                    None
+                };
```
greged93, line 727: here is an example of a case where `retry_with_backoff` could lead to issues: let's say `txn.delete_batches_gt_batch_index(prev_batch_index).await?;` doesn't error and we get `affected > 0`. Then the database has an issue at `txn.delete_l2_blocks_gt_batch_index(prev_batch_index).await?;`. Because we retry the whole task, we will try again from the start, only this time `affected == 0`, leading us to believe the safe head shouldn't be changed (while it should).

greged93 (commented):
I'm not sure `retry_with_backoff` is the strategy we want, it might be a better to have a wrapper around each provider (RPC, DB,...) that implements a retry strategy.

crates/chain-orchestrator/src/lib.rs
yiweichi: Make sense!

crates/chain-orchestrator/src/lib.rs
```diff
@@ -589,10 +615,24 @@ impl<
...
-            txn.unwind(chain_spec.genesis_hash(), l1_block_number).await?;
-        txn.commit().await?;
+            retry_with_defaults("unwind", || async {
+                let txn = database.tx_mut().await?;
+                let UnwindResult {
+                    l1_block_number,
+                    queue_index,
+                    l2_head_block_number,
+                    l2_safe_block_info,
+                } = txn.unwind(chain_spec.genesis_hash(), l1_block_number).await?;
```
greged93, line 626: I don't think we maintain idempotency between each retry for the unwind call.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -673,36 +725,43 @@ impl<
...
+            if tx.get_batch_by_index(prev_batch_index).await?.is_none() {
+                return Err(ChainOrchestratorError::BatchCommitGap(batch_clone.index))
+            }
 
-        // Perform a consistency check to ensure the previous commit batch exists in the database.
-        if tx.get_batch_by_index(prev_batch_index).await?.is_none() {
-            return Err(ChainOrchestratorError::BatchCommitGap(batch.index))
-        }
+            // remove any batches with an index greater than the previous batch.
+            let affected = tx.delete_batches_gt_batch_index(prev_batch_index).await?;
```
greged93, line 740: same issue here

crates/chain-orchestrator/src/lib.rs
```diff
@@ -712,22 +771,29 @@ impl<
...
-            tx.commit().await?;
-            return Ok(Some(ChainOrchestratorEvent::BatchFinalized(block_number, finalized_batches)))
-        }
+            // Get all unprocessed batches that have been finalized by this L1 block finalization.
+            let finalized_block_number = finalized_block_number.load(Ordering::Relaxed);
+            if finalized_block_number >= block_number {
+                let finalized_batches = tx
+                    .fetch_and_update_unprocessed_finalized_batches(finalized_block_number)
+                    .await?;
+                tx.commit().await?;
```
greged93, line 786: I'm not sure what happens here, if the `tx.commit().await?;` fails? Because we are using the same tx between each retry this might also break idempotency.

greged93 (commented):
I think we still have the previous idempotency issue.

frisitano (commented):
Looks good. I left some comments inline. I am wondering if we need to differentiate between the types of errors that are returned from functions? I think for now it's fine to leave as it is but in the future maybe we should consider error returned before retrying.

PR #335, opened 2025-09-22
feat: set initial fcs from db

### LINKED ISSUE #334: [Engine] Set fork choice state via database
In order to avoid an edge case where a block gets accepted by the EN and the RN crashes before L1 messages are marked as executed in the database, leading to an incorrect L1 message index at startup for the sequencer, we set the fork choice state at start by reading from the database.

greged93:
Sets the fcs at startup from the database instead of the provider. Avoids edge cases on the initial L1 message index for the sequencer, where an incorrect state could arise during restart from a crash.

Resolves #334

### DIFF
--- a/crates/chain-orchestrator/src/lib.rs
+++ b/crates/chain-orchestrator/src/lib.rs
@@ -275,13 +275,6 @@ impl<
                 optimistic_headers.push_front(header);
             }
 
-            // Persist the head block to the database.
-            persist_l2_head_with_retry(
-                database.clone(),
-                optimistic_headers.last().expect("chain can not be empty").into(),
-            )
-            .await?;
-
             *chain.lock().await = optimistic_headers;
             *optimistic_mode.lock().await = true;
             return Ok(ChainOrchestratorEvent::OptimisticSync(received_block));
@@ -447,13 +440,6 @@ impl<
             validate_l1_messages(&new_blocks, &database).await?;
         }
 
-        // Persist the head block to the database.
-        persist_l2_head_with_retry(
-            database.clone(),
-            new_blocks.last().expect("chain can not be empty").into(),
-        )
-        .await?;
-
         match reorg_index {
             // If this is a simple chain extension, we can just extend the in-memory chain and emit
             // a ChainExtended event.
@@ -550,6 +536,7 @@ impl<
                     .retry("update_l1_messages_from_l2_blocks", || async {
                         let tx = database.tx_mut().await?;
                         tx.update_l1_messages_from_l2_blocks(block_info.clone()).await?;
+                        tx.set_l2_head_block_info(head.block_info).await?;
                         tx.commit().await?;
                         Ok::<_, ChainOrchestratorError>(())
                     })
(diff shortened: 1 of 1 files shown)


crates/engine/src/fcs.rs
```diff
@@ -60,12 +62,22 @@ impl ForkchoiceState {
...
 
+    /// Creates a [`ForkchoiceState`] instance setting the `head`, `safe` and `finalized` hash to
+    /// the appropriate starting values by reading from the database.
+    pub async fn from_db<DB: DatabaseTransactionProvider>(db: DB) -> Option<Self> {
+        // TODO: important to add retry for these operations.
+        let tx = db.tx().await.ok()?;
+        let latest_l2_block = tx.get_l2_blocks().await.ok()?.next().await?.ok()?;
+        let (latest_safe_block, _) = tx.get_latest_safe_l2_info().await.ok()??;
+        Some(Self { head: latest_l2_block, safe: latest_safe_block, finalized: latest_safe_block })
+    }
```
frisitano, line 73: Is it also suitable to set the `safe` and `finalized` heads using the database? I wonder if we should be setting the `safe` and `finalized` using the EN provider and the `head` using the database. My concern is that we could accidentally issue an FCU containing a `finalized` head, which is older than the `finalized` head of the EN. This would be an invalid FCU. For the `safe` and `finalized` they are deterministic and therefore will be eventually consistent so I don't think it's an issue to use the EN provider to set them.

crates/node/src/args.rs
```diff
@@ -243,9 +244,30 @@ impl ScrollRollupNodeConfig {
...
-            .await
-            .unwrap_or_else(chain_spec_fcs);
+        let mut fcs =
+            ForkchoiceState::from_provider(&l2_provider).await.unwrap_or_else(chain_spec_fcs);
+
+        // Update the head block info from the database if available.
+        if let Some(latest_block) = db.tx().await?.get_l2_blocks().await?.next().await {
+            let latest_block = latest_block?;
+            fcs.update_head_block_info(latest_block);
+        }
```
frisitano, line 254: We only persist safe l2 blocks in the database. Unsafe blocks are not persisted in the database, we only persist a mapping of L1 message -> L2 block number. This may be a shortcoming of the data model that needs to be revised. We may need to revert to persisting unsafe blocks in the database as well.

frisitano: This works, but it could lead to relatively deep reorgs if L1 messages are not included in a large number of blocks. An alternative solution would be to track the current chain head in the database metadata table.

crates/manager/src/manager/mod.rs
```diff
@@ -381,26 +381,17 @@ where
...
-                trace!(target: "scroll::node::manager", head = ?chain_import.chain.last().unwrap().header.clone(), peer_id = ?chain_import.peer_id.clone(),  "Received chain extension from peer");
-                // Issue the new chain to the engine driver for processing.
-                self.engine.handle_chain_import(chain_import)
-            }
+            ChainOrchestratorEvent::ChainExtended(chain_import) |
             ChainOrchestratorEvent::ChainReorged(chain_import) => {
                 self.metrics
                     .handle_chain_import_block_number
                     .set(chain_import.chain.last().unwrap().number as f64);
-                trace!(target: "scroll::node::manager", head = ?chain_import.chain.last().unwrap().header, ?chain_import.peer_id, "Received chain reorg from peer");
```
greged93, line 396: we have logs at the start of the method, so this feels like double logging.

frisitano (approved):

PR #341, opened 2025-09-25
fix: database prepare on startup

frisitano:
# Overview
We run into an issue in which, on startup, we do not correctly clear batches and set the starting L2 safe block. This edge case occurs when multiple batches are present in the same L1 block. This led to an invalid reorg. 

Previously, we would only clear the latest batch, but when we resync the L1 and process the L1 block, it would trigger an L1 consolidation of all batches contained in the L1 block instead of just the latest, which was deleted. As such, we now delete all batches included in the L1 block on startup.

### DIFF
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -222,7 +222,6 @@ pub trait DatabaseWriteOperations: WriteConnectionProvider + DatabaseReadOperati
         };
         let batch = self.get_batch_by_index(batch_info.index).await?.expect("batch must exist");
         Ok((Some(block_info), Some(batch.block_number.saturating_add(1))))
-
     }
 
     /// Delete all L2 blocks with a block number greater than the provided block number.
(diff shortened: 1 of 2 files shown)


crates/database/db/src/operations.rs
```diff
@@ -205,25 +205,24 @@ pub trait DatabaseWriteOperations: WriteConnectionProvider + DatabaseReadOperati
...
 
-        Ok(safe)
+        match self.get_latest_safe_l2_info().await? {
+            Some((block_info, batch_info)) if block_info.number > 0 => {
+                let batch =
+                    self.get_batch_by_index(batch_info.index).await?.expect("batch must exist");
+                Ok((Some(block_info), Some(batch.block_number.saturating_add(1))))
+            }
+            _ => Ok((None, None)),
+        }
```
greged93, line 225: nit (feel free to skip):

```suggestion
        let Some((block_info, batch_info)) =
            self.get_latest_safe_l2_info().await?.filter(|(block_info, _)| block_info.number > 0)
        else {
            return Ok((None, None))
        };
        let batch = self.get_batch_by_index(batch_info.index).await?.expect("batch must exist");
        Ok((Some(block_info), Some(batch.block_number.saturating_add(1))))

```

greged93 (approved):
lgtm! left some small nits inline, feel free to merge

PR #346, opened 2025-09-30
feat: check continuity of blocks + rework wrappers

greged93:


crates/derivation-pipeline/src/lib.rs
```diff
@@ -321,7 +363,13 @@ pub async fn derive<L1P: L1Provider + Sync + Send, L2P: BlockDataProvider + Sync
...
             gas_limit: Some(block.context.gas_limit),
         };
-        attributes.push(attribute);
+        attributes.push(WithL1FinalizedBlockNumber::new(
+            batch.block_number,
+            WithL2BlockNumber::new(
+                block.context.number,
+                WithBatchInfo::new(batch.index, batch.hash, attribute),
+            ),
+        ));
```
greged93, line 372: The update adds a little work when initializing the structures, but the information attached to the attribute is more explicit.

crates/engine/src/future/mod.rs
```diff
@@ -267,16 +267,19 @@ async fn handle_payload_attributes<EC, P>(
...
 {
     tracing::trace!(target: "scroll::engine::future", ?fcs, ?payload_attributes_with_batch_info, "handling payload attributes");
 
-    let ScrollPayloadAttributesWithBatchInfo { mut payload_attributes, batch_info } =
-        payload_attributes_with_batch_info.inner.clone();
+    let batch_info = BatchInfo::new(
+        payload_attributes_with_batch_info.index,
+        payload_attributes_with_batch_info.hash,
+    );
+    let mut payload_attributes = payload_attributes_with_batch_info.inner.inner.inner.clone();
```
frisitano, line 282: This type is getting unwieldy with the requirement of `.inner.inner.inner`. Should we consider refactoring the base types to include the additional metadata?

crates/primitives/src/block.rs
```diff
@@ -83,69 +84,74 @@ impl arbitrary::Arbitrary<'_> for BlockInfo {
...
-
-    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let block_number = self.number;
-        let index = self.index;
-        let inner = ready!(Pin::new(&mut self.get_mut().inner).poll(cx));
-        Poll::Ready(WithBatchInfo::new(index, block_number, inner))
-    }
-}
+/// Type alias for a wrapper type with the full L2 metadata.
+pub type WithFullL2Meta<T> = WithL1FinalizedBlockNumber<WithL2BlockNumber<WithBatchInfo<T>>>;
```
frisitano, line 154: Same comment as above, this is a complex type. Should we push the additional metadata into the inner type `T`.

frisitano (commented):
Left some comments inline.

frisitano:
@greged93 can we close this and open a new implementation based on the new derivation pipeline pattern?

PR #350, opened 2025-10-06
feat: bench derivation pipeline

greged93:
Introduces a derivation pipeline benchmark running with network calls to the AWS S3 blob storage. Also runs the benchmark in CI using codspeed.


jonastheis (dismissed):
lgtm. just need to fix test and lint

jonastheis (approved):

PR #358, opened 2025-10-13
feat: update retry mechanism

greged93:
closes: #319

### DIFF
--- a/crates/node/src/args.rs
+++ b/crates/node/src/args.rs
@@ -270,7 +270,6 @@ impl ScrollRollupNodeConfig {
                 let l2_head_block_number = tx.get_l2_head_block_number().await?;
                 tx.purge_l1_message_to_l2_block_mappings(Some(l2_head_block_number + 1)).await?;
 
-                let l2_head_block_number = tx.get_l2_head_block_number().await?;
                 Ok::<_, DatabaseError>((l1_start_block_number, l2_head_block_number))
             })
             .await?;
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -928,7 +928,7 @@ impl<T: ReadConnectionProvider + Sync + ?Sized> DatabaseReadOperations for T {
         n: usize,
     ) -> Result<Vec<BlockDataHint>, DatabaseError> {
         Ok(models::block_data::Entity::find()
-            .filter(models::block_data::Column::Number.eq(block_number as i64))
+            .filter(models::block_data::Column::Number.gte(block_number as i64))
             .limit(Some(n as u64))
             .all(self.get_connection())
             .await?


crates/chain-orchestrator/src/lib.rs
```diff
@@ -327,10 +332,15 @@ impl<
                 let _ = tx.send(self.event_listener());
             }
             ChainOrchestratorCommand::Status(tx) => {
-                let db_tx = self.database.tx().await?;
-                let l1_latest = db_tx.get_latest_l1_block_number().await?;
-                let l1_finalized = db_tx.get_finalized_l1_block_number().await?;
-                let l1_processed = db_tx.get_processed_l1_block_number().await?;
+                let (l1_latest, l1_finalized, l1_processed) = self
+                    .database
+                    .tx_mut(|tx| async move {
```
frisitano:
```suggestion
                    .tx(|tx| async move {
```

crates/chain-orchestrator/src/lib.rs
```diff
@@ -478,9 +490,13 @@ impl<
...
-        let tx = self.database.tx_mut().await?;
-        tx.insert_batch_consolidation_outcome(batch_consolidation_outcome.clone()).await?;
-        tx.commit().await?;
+        let consolidation_outcome = batch_consolidation_outcome.clone();
+        self.database
+            .tx_mut(move |tx| {
+                let consolidation_outcome = consolidation_outcome.clone();
+                async move { tx.insert_batch_consolidation_outcome(consolidation_outcome).await }
+            })
+            .await?;
```
frisitano:
```suggestion
        // Insert the batch consolidation outcome into the database.
        self.database
            .insert_batch_consolidation_outcome(batch_consolidation_outcome.clone())
            .await?;
```

crates/chain-orchestrator/src/lib.rs
```diff
@@ -833,10 +839,12 @@ impl<
         }
 
         // We optimistically persist the signature upon passing consensus checks.
-        let tx = self.database.tx_mut().await?;
-        tx.insert_signature(block_with_peer.block.header.hash_slow(), block_with_peer.signature)
+        let block_hash = block_with_peer.block.header.hash_slow();
+        self.database
+            .tx_mut(move |tx| async move {
+                tx.insert_signature(block_hash, block_with_peer.signature).await
+            })
```
frisitano: no need for tx_mut, use the method directly.

crates/chain-orchestrator/src/lib.rs
```diff
@@ -1137,12 +1150,17 @@ impl<
...
+            .tx_mut(move |tx| {
+                let blocks_to_validate = blocks_to_validate.clone();
+                async move {
+                    tx.update_l1_messages_from_l2_blocks(
+                        blocks_to_validate.into_iter().map(|b| (&b).into()).collect(),
+                    )
+                    .await
+                }
+            })
+            .await?;
```
frisitano: use the method directly

crates/chain-orchestrator/src/lib.rs
```diff
@@ -1241,17 +1256,17 @@ async fn compute_l1_message_queue_hash(
...
     } else if l1_message.queue_index > l1_v2_message_queue_start_index {
         let index = l1_message.queue_index - 1;
-        let mut input = Retry::default()
-            .retry("get_l1_message_by_index", || async {
-                let tx = database.tx().await?;
-                let input = tx.get_l1_message_by_index(index).await?;
-                Ok::<_, ChainOrchestratorError>(input)
+        let mut input = database
+            .tx(move |tx| async move {
+                tx.get_n_l1_messages(Some(L1MessageKey::from_queue_index(index)), 1).await
```
frisitano: use the method directly.

crates/database/db/src/service/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+                let res = f(tx.clone()).await;
+
+                // The `WriteQuery` cannot clone the atomic reference to the transaction, or the
+                // below will fail, and we won't be able to commit/rollback the transaction.
+                let tx = Arc::try_unwrap(tx);
+
+                if res.is_ok() {
+                    tx.map_err(|_| DatabaseError::CommitFailed)?.commit().await?;
+                } else {
+                    tx.map_err(|_| DatabaseError::RollbackFailed)?.rollback().await?;
```
frisitano: nit, should we execute `.commit()` / `.rollback()` before we `map_err(..)`? No particular difference but it's more readable. What do you think?

crates/node/src/args.rs
```diff
@@ -263,15 +260,13 @@ impl ScrollRollupNodeConfig {
...
-        let l2_head_block_number = tx.get_l2_head_block_number().await?;
-        tx.purge_l1_message_to_l2_block_mappings(Some(l2_head_block_number + 1)).await?;
-        tx.commit().await?;
+            db.prepare_on_startup(chain_spec.genesis_hash()).await?;
+        let l2_head_block_number = db.get_l2_head_block_number().await?;
+        db.purge_l1_message_to_l2_block_mappings(Some(l2_head_block_number + 1)).await?;
 
         // Update the head block info if available and ahead of finalized.
-        let l2_head_block_number = db.tx().await?.get_l2_head_block_number().await?;
+        let l2_head_block_number = db.get_l2_head_block_number().await?;
```
frisitano: should we combine this with the transaction above?

frisitano (commented):
Looks great! Added some minor comments inline, primiarly about when we should use the raw methods and when to instantiate a transaction.

frisitano (dismissed):
lgtm

greged93:
@frisitano should we merge this with the old benchmark and fix in another PR?

frisitano:
> @frisitano should we merge this with the old benchmark and fix in another PR?

Yes lets revent back to the old bencmarks, get CI green and merge. We can look at codespeed in another PR as you suggest.

frisitano:
> @frisitano regression seems to have came from not having a `get_n_block_data_hint`. With the added method in [22e5d4c](https://github.com/scroll-tech/rollup-node/pull/358/commits/22e5d4c851ea199d13df5027bbaaa4d1250e3caf), regression is corrected.

Can you elaborate on this please? In main we don't have a `get_n_block_data_hint` so why would this result in a regression?

frisitano (approved):
Looks good. lgtm

greged93:
> edit: Is this because in the derivation pipeline using the new api we would create a new transaction for each block data hint we fetched?

Yes, it seems like this slows the derivation process.

frisitano (approved):
lgtm

PR #367, opened 2025-10-16
feat: update metrics

greged93:


crates/chain-orchestrator/src/lib.rs
```diff
@@ -300,6 +302,7 @@ impl<
                         .as_mut()
                         .expect("signer must be present")
                         .sign_block(block.clone())?;
+                    self.metric_handler.finish_block_building_recording();
```
frisitano: if we don't allow empty blocks then the `finalize` method below will return `None`. We should handle that case.
```rust
self
                    .sequencer
                    .as_mut()
                    .expect("sequencer must be present")
                    .finalize_payload_building(payload_id, &mut self.engine)
```

frisitano: On second thought, maybe it was your intention to skip it such that we don't report metrics for empty blocks?

frisitano: I think these could both be valuable to have

PR #377, opened 2025-10-20
feat(watcher): Configureable block query range

### LINKED ISSUE #374: [Watcher] Configureable block query range
# Overview
We should allow the block query range to be configureable via cli such that users can specify the range of blocks to query. This will allow users on free tiers of rpc provider package to consolidate with L1 (potentailly at a lower performance).

https://github.com/scroll-tech/rollup-node/blob/4b1d4d5f45932a0d220894a913616185db4da482/crates/watcher/src/lib.rs#L32-L33

varun-doshi:
Fixes #374

### DIFF
--- a/crates/watcher/src/lib.rs
+++ b/crates/watcher/src/lib.rs
@@ -639,7 +639,8 @@ where
 
     /// Returns the next range of logs, for the block range in
     /// \[[`current_block`](field@L1Watcher::current_block_number);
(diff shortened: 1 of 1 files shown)


varun-doshi:
> @varun-doshi CI is failing

Ahh the errors were not showing for me on `make pr`...had to update my rust version

frisitano (approved):
lgtm

PR #378, opened 2025-10-20
feat(rpc): Modify getL1Message* rpc to acces L1MesageKey

### LINKED ISSUE #372: [RPC] Modify `getL1Message*` rpc to acces `L1MesageKey`
> should we directly implement `get_l1_message` and pass a L1MessageKey? 

 _Originally posted by @greged93 in [#370](https://github.com/scroll-tech/rollup-node/pull/370/files#r2444878531)_

varun-doshi:
Fixes #372 

- `get_l1_message_by_index` now takes a `L1MessageKey` instead of an `index: u64`
- `GetL1MessageByIndex` under `DatabaseQuery` takes a `L1MessageKey`

### DIFF
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -58,6 +58,9 @@ pub trait DatabaseWriteOperations {
     /// Insert an [`L1MessageEnvelope`] into the database.
     async fn insert_l1_message(&self, l1_message: L1MessageEnvelope) -> Result<(), DatabaseError>;
 
+    /// Sets the `skipped` column to true for the provided list of L1 messages queue indexes.
+    async fn update_skipped_l1_messages(&self, indexes: Vec<u64>) -> Result<(), DatabaseError>;
+
     /// Delete all [`L1MessageEnvelope`]s with a block number greater than the provided block
     /// number and return them.
     async fn delete_l1_messages_gt(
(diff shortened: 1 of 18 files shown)


crates/node/src/add_ons/rpc.rs
```diff
@@ -83,7 +84,10 @@ pub trait RollupNodeExtApi {
 
     /// Returns the L1 message by index.
     #[method(name = "getL1MessageByIndex")]
-    async fn get_l1_message_by_index(&self, index: u64) -> RpcResult<Option<L1MessageEnvelope>>;
+    async fn get_l1_message_by_index(
+        &self,
+        l1_message_key: L1MessageKey,
+    ) -> RpcResult<Option<L1MessageEnvelope>>;
```
frisitano: I think we should leave the previous `get_l1_message_by_index` rpc method and introduce a new method `get_l1_message_by_key`.

crates/database/db/src/operations.rs
```diff
@@ -1040,7 +1040,7 @@ impl<T: ReadConnectionProvider + Sync + ?Sized> DatabaseReadOperations for T {
 /// A key for an L1 message stored in the database.
 ///
 /// It can either be the queue index, queue hash or the transaction hash.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum L1MessageKey {
```
frisitano: Lets use camel case for serde

crates/database/db/src/operations.rs
frisitano: add `#[serde(rename = "notIncludedBlockDepth")]`

frisitano (changes requested):
Thanks for your contribution! I've added a few comments inline. Additionally, it would be great if we could add a few tests:
- One round trip of any vairant (serialize and then deserialze)
- One or two hand crafted serialized instances and then deserialize to ensure we can construct the correct enum variant. This gives us confidence in the rpc API we can expect.

crates/database/db/src/operations.rs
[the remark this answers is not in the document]
frisitano: ahhh ok, no that was my mistake. Let's leave as is.

frisitano (approved):
thanks! lgtm

Issue #369, opened 2025-10-17
[L1 Wather] Gracefully handle genesis batch commit event


yiweichi:
For now when RN receives a genesis batch commit event(batchIndex=0) from L1, there will be a underflow issue happens [here](https://github.com/scroll-tech/rollup-node/blob/50b63b25a540a94be8f083479f63a2663c0048c7/crates/chain-orchestrator/src/lib.rs#L658),  and further if we skip the batch 0, we will have error when receiving the next batch(batch 0) [here](https://github.com/scroll-tech/rollup-node/blob/50b63b25a540a94be8f083479f63a2663c0048c7/crates/chain-orchestrator/src/lib.rs#L663).
We handle this on Sepolia by set a sync start block after we commit genesis batch and manually insert batch 0 to database of RN, but we might as well to have a more gracefully way to handle it for the convenience of other project (Cloak, DogeOS).

PR #383, opened 2025-10-22
feat: gracefully handle commit/finalize genesis batch event

yiweichi:
This PR add feature to gracefully handle commit/finalize genesis batch event. 

- set the mainnet/sepolia start block number to the one we commit genesis batch
- skip event commit/finalize genesis batch, since it was already persisted default empty batch in db.

corresponding issue: https://github.com/scroll-tech/rollup-node/issues/369

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11627,6 +11627,7 @@ dependencies = [
  "sea-orm",
  "serde",
  "serde_json",
+ "strum 0.27.2",
  "tempfile",
  "thiserror 2.0.16",
  "tokio",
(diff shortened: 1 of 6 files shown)


frisitano (commented):
I'm a little concerned that we are adding additional computational overhead in the statement for no functional difference in processing:
```
                    .filter(|decoded| !decoded.data.batch_index.is_zero())
```

yiweichi:
> I'm a little concerned that we are adding additional computational overhead in the statement for no functional difference in processing:
> 
> ```
>                     .filter(|decoded| !decoded.data.batch_index.is_zero())
> ```
Claude says the computational overhead is minimal, shouldn't bring performance issue for L1 watcher at all. And we kinda need this filter to skip genesis batch commit event.

I can move the filter logic above `notifications.push(L1Notification::BatchCommit(BatchCommitData {`, but this look more elegant.

frisitano (approved):
lgtm

PR #387, opened 2025-10-23
fix: startup latest block

frisitano:
# Overview
This PR addresses the pattern we use to determine the latest startup block. This is achieved by looping from the head block back towards to finalized block and selecting the latest block that is present in the exectuion node.


greged93 (approved):
lgtm!

Issue #381, opened 2025-10-21
[Database] L1 message `L1MessageKey::BlockNumber` lookup edge case


frisitano:
> We should consider the case where the last message in a batch is skipped and update the `L1MessageKey::BlockNumber` query to account for it. 

 _Originally posted by @frisitano in [#380](https://github.com/scroll-tech/rollup-node/pull/380/files#r2448068346)_

In practice, this should never happen as we no longer skip L1 messages and we never had a skipped L1 message at the end of a batch historically. However, in the name of being comprehensive it would be good to conver this edge case.

yiweichi:
Close this by PR https://github.com/scroll-tech/rollup-node/pull/402

PR #402, opened 2025-10-29
fix: L1 message `L1MessageKey::BlockNumber` lookup edge case

yiweichi:
This PR handle the edge case that the last message in a batch is skipped and update the L1MessageKey::BlockNumber query to account for it. We filter out the skipped message during the query.

Corresponding issue: https://github.com/scroll-tech/rollup-node/issues/381

### DIFF
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -857,12 +857,12 @@ impl<T: ReadConnectionProvider + Sync + ?Sized> DatabaseReadOperations for T {
                     // Yield n messages starting from the found queue index + 1. Only return
                     // messages that have not been skipped (skipped = false) to handle the edge
                     // case where the last message in a batch is skipped.
+                    let condition = Condition::all()
+                        // We add 1 to the queue index to constrain across block boundaries
+                        .add(models::l1_message::Column::QueueIndex.gte(record.queue_index + 1))
+                        .add(models::l1_message::Column::Skipped.eq(false));
                     Ok(models::l1_message::Entity::find()
-                        .filter(
-                            // We add 1 to the queue index to constrain across block boundaries
-                            models::l1_message::Column::QueueIndex.gte(record.queue_index + 1),
-                        )
-                        .filter(models::l1_message::Column::Skipped.eq(false))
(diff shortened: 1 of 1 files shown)


crates/database/db/src/operations.rs
```diff
@@ -854,12 +854,15 @@ impl<T: ReadConnectionProvider + Sync + ?Sized> DatabaseReadOperations for T {
...
-                    // Yield n messages starting from the found queue index + 1.
+                    // Yield n messages starting from the found queue index + 1. Only return
+                    // messages that have not been skipped (skipped = false) to handle the edge
+                    // case where the last message in a batch is skipped.
                     Ok(models::l1_message::Entity::find()
                         .filter(
                             // We add 1 to the queue index to constrain across block boundaries
                             models::l1_message::Column::QueueIndex.gte(record.queue_index + 1),
                         )
+                        .filter(models::l1_message::Column::Skipped.eq(false))
```
frisitano: can we combine this into the filter above using a pattern like:
```rust
                let condition = Condition::all()
                    .add(models::l1_message::Column::QueueIndex.gte(record.queue_index + 1))
                    .add(models::l1_message::Column::Skipped.eq(false));
                // Yield n messages matching the condition ordered by increasing queue index.
                Ok(models::l1_message::Entity::find()
                    .filter(condition)
```

crates/database/db/src/operations.rs
```diff
@@ -854,12 +854,15 @@ impl<T: ReadConnectionProvider + Sync + ?Sized> DatabaseReadOperations for T {
                     .one(self.get_connection())
                     .await?
                 {
-                    // Yield n messages starting from the found queue index + 1.
+                    // Yield n messages starting from the found queue index + 1. Only return
```
frisitano: Do we need to add a filter on `Skipped` for the else branch below as well?

yiweichi: Nice catch! an edge case of the edge case. fixed by commit [a6bef19](https://github.com/scroll-tech/rollup-node/pull/402/commits/a6bef194e6ebae4790b1b1cf4f2583b7392dc2cf)

Issue #320, opened 2025-09-16
[Sequencer] sequence on top of older block feature (similar to `debug_setHead`)


jonastheis:
In case there's some unexpected issue and we need to revert some L2 blocks we should have the option to reset the node to a certain block height. In l2geth this is can be done via `debug_setHead`.

Issue #352, opened 2025-10-08
[Chain Orchestrator] Improve `UpdateFcsHead` method to add L2 transactions to transaction pool


frisitano:
# Overview
Currently, we implement a relatively naive solution for the `UpdateFcsHead` rpc method. This involves two steps:

1. We purge all mappings of L1 messages `->` l2 blocks (effectively marking them suitable for inclusion in future blocks.
2. Updating the head in the fork choice state.

@Thegaram proposed a more elegant solution in which we extract all L2 messages from the reverted chain and push them into the transaction pool such that they can be included in future blocks.

yiweichi:
Close this by PR https://github.com/scroll-tech/rollup-node/pull/406

PR #406, opened 2025-11-02
feat: improve UpdateFcsHead to add reverted transations to tx pool

yiweichi:
This PR enhances L2 block reorg handling. When a reorg occurs, we:
1. Identify all transactions from reverted L2 blocks 
2. Update forkchoice to the new canonical chain 
3. Re-submit the reverted transactions to the transaction pool using send_raw_transaction
This ensures that reverted but valid transactions are not lost and may be included in future blocks.

Corresponding issue: https://github.com/scroll-tech/rollup-node/issues/352
closes: #320, #352

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -9,7 +9,7 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: peaceiris/actions-mdbook@v2
         with:
           mdbook-version: 'latest'


crates/chain-orchestrator/src/lib.rs
```diff
@@ -586,9 +588,26 @@ impl<
...
+                for number in (block_number + 1)..=current_head_number {
+                    let block = self
+                        .l2_client
+                        .get_block_by_number(number.into())
+                        .full()
+                        .await?
+                        .ok_or_else(|| ChainOrchestratorError::L2BlockNotFoundInL2Client(number))?;
+
+                    let block = block.into_consensus().map_transactions(|tx| tx.inner.into_inner());
+                    reverted_transactions.extend(block.body.transactions().cloned());
```
frisitano: Instead of using `.cloned()`, can we consume the transactions from the block so there is no cloning required? Additionally, I think we can add the filtering of L1 messages here using `.filter()`.

frisitano (commented):
Looks good. Added some comments inline.

frisitano:
Can we also update the logic for the `UpdateFcsHead`:
https://github.com/scroll-tech/rollup-node/blob/e9cff2de58951fa00491fc25d3cb97c7e1186f44/crates/chain-orchestrator/src/lib.rs#L358-L368

frisitano (approved):
lgtm

PR #407, opened 2025-11-03
tests: add test fixture for easier abstraction

greged93:
PR for the update of the test setup. Provides a set of high level abstractions that allow testers to easily set up initial tests conditions, run the test and directly expect test outcome.

Below are some examples of the usage of the framework:
```rust
    // Start a sequencer node with a follower node and configure the sequencer.
    let mut fixture = TestFixture::builder()
     .sequencer()
     .followers(1)
     .auto_start(true)
     .allow_empty_blocks(false)
     .build()
     .await?;

    // Set the l1 to synced and expect an Optimistic Sync event.
    fixture.l1().sync().await?;
    fixture.expect_event().optimistic_sync().await?;
   
    // Inject a transfer to the sequencer and build a block, expecting block number 10, containing the injected transaction.
    let hash = fixture.inject_transfer().await?;
    fixture.build_block().expect_block_number(10).expect_tx(hash).await_block().await?;

    // For node 0 (sequencer node), add a L1 message with L1 block number 10, a sender, value 0, and destination 0.
    fixture
        .l1()
        .for_node(0)
        .add_message()
        .at_block(10)
        .sender(address!("f39Fd6e51aad88F6F4ce6aB8827279cffFb92266"))
        .value(0)
        .to(Address::ZERO)
        .send()
        .await?;
```

### DIFF
--- a/crates/codec/src/lib.rs
+++ b/crates/codec/src/lib.rs
@@ -41,7 +41,7 @@ pub enum Codec {
 
 impl Codec {
     /// Decodes the input data and returns the decoded [`Batch`].
-    pub fn decode<T: CommitDataSource>(input: &T) -> Result<Batch, CodecError> {
+    pub fn decode<T: CommitDataSource>(input: T) -> Result<Batch, CodecError> {
         let calldata = input.calldata();
         let version = get_codec_version(calldata)?;
 
--- a/crates/derivation-pipeline/src/data_source.rs
+++ b/crates/derivation-pipeline/src/data_source.rs
@@ -1,19 +1,20 @@
-use alloy_eips::eip4844::Blob;
 use alloy_primitives::Bytes;
 use scroll_codec::CommitDataSource;
 
 /// Holds the data for the codec.
-pub(crate) struct CodecDataSource<'a> {
-    pub(crate) calldata: &'a Bytes,
-    pub(crate) blob: Option<&'a Blob>,
+pub(crate) struct CodecDataSource<Calldata, Blob> {
+    pub(crate) calldata: Calldata,
+    pub(crate) blob: Option<Blob>,
 }
 
-impl<'a> CommitDataSource for CodecDataSource<'a> {
+impl<Calldata: AsRef<Bytes>, Blob: AsRef<alloy_eips::eip4844::Blob>> CommitDataSource
+    for CodecDataSource<Calldata, Blob>
+{
     fn calldata(&self) -> &Bytes {
-        self.calldata
+        self.calldata.as_ref()
     }
 
-    fn blob(&self) -> Option<&Blob> {
-        self.blob
+    fn blob(&self) -> Option<&alloy_eips::eip4844::Blob> {
+        self.blob.as_ref().map(|b| b.as_ref())
     }
 }
--- a/crates/derivation-pipeline/src/lib.rs
+++ b/crates/derivation-pipeline/src/lib.rs
@@ -10,7 +10,7 @@ use futures::{stream::FuturesOrdered, Stream, StreamExt};
 use rollup_node_primitives::{BatchCommitData, BatchInfo, BatchStatus, L1MessageEnvelope};
 use rollup_node_providers::L1Provider;
 use scroll_alloy_rpc_types_engine::{BlockDataHint, ScrollPayloadAttributes};
-use scroll_codec::{decoding::payload::PayloadData, Codec};
+use scroll_codec::{decoding::payload::PayloadData, Codec, CodecError, DecodingError};
 use scroll_db::{Database, DatabaseReadOperations, L1MessageKey};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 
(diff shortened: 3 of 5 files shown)


crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+pub struct EventWaiter<'a> {
+    fixture: &'a mut TestFixture,
+    node_index: usize,
+    timeout_duration: Duration,
+}
+
+impl<'a> EventWaiter<'a> {
+    /// Create a new event waiter.
+    pub fn new(fixture: &'a mut TestFixture, node_index: usize) -> Self {
+        Self { fixture, node_index, timeout_duration: Duration::from_secs(30) }
```
frisitano: Let's move the default timeout to a constant value at the top of the file after imports.

crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    /// Set a custom timeout for waiting.
+    pub const fn timeout(mut self, duration: Duration) -> Self {
+        self.timeout_duration = duration;
+        self
+    }
+
+    /// Wait for a block sequenced event.
+    pub async fn block_sequenced(self, target: u64) -> eyre::Result<ScrollBlock> {
```
frisitano:
```suggestion
    pub async fn block_sequenced(self, target_block_number: u64) -> eyre::Result<ScrollBlock> {
```

crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+                Some((*from, *to))
+            } else {
+                None
+            }
+        })
+        .await
+    }
+
+    /// Wait for a chain extended event.
+    pub async fn chain_extended(self, target: u64) -> eyre::Result<()> {
```
frisitano:
```suggestion
    pub async fn chain_extended(self, target_block_number: u64) -> eyre::Result<()> {
```

crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// Wait for any event and extract a value from it.
+    pub async fn extract<T>(
+        self,
+        mut extractor: impl FnMut(&ChainOrchestratorEvent) -> Option<T>,
+    ) -> eyre::Result<T> {
+        self.wait_for_event(move |e| extractor(e)).await
+    }
+
+    /// Wait for N events matching a predicate.
+    pub async fn where_event_n(
```
frisitano:
```suggestion
    pub async fn where_n_events(
```

crates/node/src/test_utils/network_helpers.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Create a new network helper for a specific node.
+    pub const fn new(fixture: &'a TestFixture, node_index: usize) -> Self {
+        Self { fixture, node_index }
+    }
+
+    /// Get the network handle for this node.
+    pub async fn network_handle(
+        &self,
+    ) -> eyre::Result<scroll_network::ScrollNetworkHandle<ScrollNetworkHandle>> {
+        self.fixture.nodes[self.node_index]
```
frisitano: Is there a reason we don't have this in the `TestFixture` when we construct it?

crates/node/src/test_utils/block_builder.rs
```diff
@@ -0,0 +1,171 @@
...
+use reth_primitives_traits::transaction::TxHashRef;
+use reth_scroll_primitives::ScrollBlock;
+use scroll_alloy_consensus::ScrollTransaction;
+
+/// Builder for constructing and validating blocks in tests.
+#[derive(Debug)]
+pub struct BlockBuilder<'a> {
+    fixture: &'a mut TestFixture,
+    expected_tx_hashes: Vec<B256>,
+    expected_tx_count: Option<usize>,
```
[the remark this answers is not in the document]
greged93: I think that could happen yes, where you want to assert the tx count and just verify a single tx. Let me know if you wish for me to update this or leave as is.

crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+                Some(block_with_peer.block.clone())
+            } else {
+                None
+            }
+        })
+        .await
+    }
+
+    /// Wait for any event where the predicate returns true.
+    pub async fn where_event(
```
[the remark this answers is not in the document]
greged93: Instead of having to worry about returning the event, you can just send a closure that returns true or false for for the event, this is just to have a nicer interface but can remove.

crates/node/src/test_utils/event_utils.rs
```diff
@@ -0,0 +1,442 @@
...
+
+use futures::{FutureExt, StreamExt};
+use reth_scroll_primitives::ScrollBlock;
+use rollup_node_chain_orchestrator::ChainOrchestratorEvent;
+use rollup_node_primitives::ChainImport;
+use tokio::time::timeout;
+
+/// Builder for waiting for specific events with optional assertions.
+#[derive(Debug)]
+pub struct EventWaiter<'a> {
```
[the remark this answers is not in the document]
greged93: hmmmm I think that might make sense yes

crates/node/src/test_utils/network_helpers.rs
greged93: but the `TestFixture` holds all nodes, so you would have difference network handles for each right?

PR #414, opened 2025-11-10
feat: implement improvements

greged93:
- Adds a sanity check for the length of notifications extracted from L1 logs.
- Runs the Codec decode method on  a blocking tokio task.


crates/derivation-pipeline/src/lib.rs
```diff
@@ -280,8 +280,15 @@ pub async fn derive<L1P: L1Provider + Sync + Send, DB: DatabaseReadOperations>(
...
         None
     };
-    let data = CodecDataSource { calldata: batch.calldata.as_ref(), blob: blob.as_deref() };
-    let decoded = Codec::decode(&data)?;
+
+    let data = CodecDataSource { calldata: batch.calldata.clone(), blob };
+
+    let decoded =
+        tokio::task::spawn_blocking(move || Codec::decode(data)).await.map_err(|err| {
+            DerivationPipelineError::Codec(CodecError::Decoding(DecodingError::Other(Box::new(
```
jonastheis: Why do we need to `Box` the error here?

jonastheis (approved):

crates/derivation-pipeline/src/lib.rs
frisitano: Because we use dynamic dispatch to allow for any error type that implements the trait `std::error::Error` ( [reference](https://doc.rust-lang.org/1.8.0/book/trait-objects.html)).
https://github.com/scroll-tech/rollup-node/blob/752c294cc3eceacd4cb1f44ec3c622aeb04f386f/crates/codec/src/error.rs#L35-L36

PR #428, opened 2025-11-18
test: add test cases for handling l1 events and l1 reorg

### LINKED ISSUE #420: [Testing] Test multi-mode L1
# Overview
In #409, we introduced a mechanism that allows for events from L1 to be consumed eagerly. This is to say that messages will be consumed before they are finalized and we will react to reorgs on L1 if they are to occur. We need to add more comprehensive testing to cover the following cases:
- Correct behaviour when consuming events from L1:

| Event | Sync State | Expected Outcome |
|----------|----------|----------|
| BatchCommit| Syncing | There should be no change as we only trigger batch commit events when the BatchFinalized event has been finalized on L1 | 
| BatchCommit | Synced | There should be an update to the safe head of the node |
| BatchFinalized | Syncing | This should trigger all unprocessed BatchCommit events up to the finalized batch (the BatchFinalized event must be finalized on L1). This should update both the safe and finalized head to the latest block in the last BatchCommit event |
| BatchedFinalized | Synced | This should update the finalized head (the BatchFinalized event must be finalized on L1 | 
| BatchRevert | Syncing | This should have no effect |
| BatchRevert | Synced | This should update the safe head to the last block of the batch commit before the reverted batch |
| BatchRevertRange | Syncing | This should have no effect |
| BatchRevetRange| Synced | This should update the safe head to the last block of the batch commit before the reverted batch range |

- L1 reorg of `BatchCommit`, `BatchFinalization`, `BatchCommit`, and `BatchRevert` events. It should be noted that when the `rollup-node` is still actively syncing from L1, the `rollup-node` will wait for `BatchFinalization` before it processes a batch. After the `rollup-node` finishes historical sync, as indicated by the `L1Synced` event from the watcher, it will process the `BatchCommit` event, immediately updating the safe block number.

| Reorged Event | Expected Outcome |
|----------|----------|
| BatchReverted  | Update the safe head to the last block number associated with the reverted batch  |
| BatchRangeReverted  | Update the safe head to the last block associated with the last batch of the reverted batch range  |
| BatchCommit | Update the safe head to the last block of the previous BatchCommit |
| BatchFinalized | No change because we only update the finalized head after the BatchFinalized event is finalized on L1 meaning it can never be reorged | 

- When a node shuts down and then starts up again, we should make sure that it can handle any reorgs that have happened whilst the node has been shut down. If we had unfinalized L1 events in the database when he node shutdown we should make sure these are reorged away on startup if a reorg occurred on L1.

# Implementation 
We should create batches and chains that allow us to test all the cases above. We should introduce an instance of anvil into the [`TestFixture`](https://github.com/scroll-tech/rollup-node/blob/a94f09b0ece452c6d142eee05fe578a7caf0d8e2/crates/node/src/test_utils/fixture.rs#L42-L53) introduced in #407 - anvil reference: https://github.com/foundry-rs/foundry/blob/12be76ff47ae0f4fe1a18b74a47e5f0611bab5a0/crates/anvil/src/lib.rs#L114. We should configure this anvil instance using the [anvil state](https://github.com/scroll-tech/rollup-node/blob/main/tests/anvil_state.json) we use for `docker-compose` testing.

yiweichi:
This PR adds test cases for handling l1 events and l1 reorg.

- added test cases to test the behavior of RN when handling `BatchCommit`, `BatchFinalized` and `BatchRevertRange` events in different sync mode.
- added test cases to test the behavior of RN when handling L1 block reorg with `BatchCommit`, `BatchFinalized` and `BatchRevertRange` events.

Fixes #420, #141

### DIFF
--- a/crates/watcher/src/lib.rs
+++ b/crates/watcher/src/lib.rs
@@ -337,20 +337,19 @@ where
             if self.is_synced {
                 tokio::time::sleep(SLOW_SYNC_INTERVAL).await;
             } else if self.current_block_number == self.l1_state.head {
+                // In test mode, skip notification if flag is set
+                #[cfg(feature = "test-utils")]
+                if self.test_mode_skip_synced_notification {
+                    self.is_synced = true;
+                    continue;
+                }
+
                 // if we have synced to the head of the L1, notify the channel and set the
                 // `is_synced`` flag.
-                #[cfg(feature = "test-utils")]
-                let should_skip = self.test_mode_skip_synced_notification;
-                #[cfg(not(feature = "test-utils"))]
-                let should_skip = false;
-
-                if !should_skip {
-                    if let Err(L1WatcherError::SendError(_)) =
-                        self.notify(L1Notification::Synced).await
-                    {
-                        tracing::warn!(target: "scroll::watcher", "L1 watcher channel closed, stopping the watcher");
-                        break;
-                    }
+                if let Err(L1WatcherError::SendError(_)) = self.notify(L1Notification::Synced).await
+                {
+                    tracing::warn!(target: "scroll::watcher", "L1 watcher channel closed, stopping the watcher");
+                    break;
                 }
                 self.is_synced = true;
             }
--- a/crates/database/db/src/operations.rs
+++ b/crates/database/db/src/operations.rs
@@ -831,9 +831,9 @@ impl<T: WriteConnectionProvider + ?Sized + Sync> DatabaseWriteOperations for T {
         &self,
         outcome: BatchConsolidationOutcome,
     ) -> Result<(), DatabaseError> {
-        let l2_head_block_number = outcome.l2_head_updated.then(|| {
-            outcome.blocks.last().expect("must have at least one block").block_info.number
-        });
+        if let Some(block_info) = outcome.updated_l2_head() {
+            self.set_l2_head_block_number(block_info.number).await?;
+        }
         self.insert_blocks(
             outcome.blocks.iter().map(|b| b.block_info).collect(),
             outcome.batch_info,
@@ -842,9 +842,6 @@ impl<T: WriteConnectionProvider + ?Sized + Sync> DatabaseWriteOperations for T {
         self.update_l1_messages_with_l2_blocks(outcome.blocks).await?;
         self.update_skipped_l1_messages(outcome.skipped_l1_messages).await?;
         self.update_batch_status(outcome.batch_info.hash, outcome.target_status).await?;
-        if let Some(block_number) = l2_head_block_number {
-            self.set_l2_head_block_number(block_number).await?;
-        }
         Ok(())
     }
 
--- a/crates/node/src/test_utils/event_utils.rs
+++ b/crates/node/src/test_utils/event_utils.rs
@@ -179,8 +179,8 @@ impl<'a> EventWaiter<'a> {
         Ok(())
     }
 
-    /// Wait for L1 block finalized event on all specified nodes.
-    pub async fn l1_block_finalized(self) -> eyre::Result<()> {
+    /// Wait for L1 genesis block finalized event on all specified nodes.
+    pub async fn l1_genesis_finalized(self) -> eyre::Result<()> {
         self.l1_block_finalized_at_least(0).await
     }
 
--- a/crates/node/tests/e2e.rs
+++ b/crates/node/tests/e2e.rs
@@ -1266,7 +1266,7 @@ async fn consolidates_committed_batches_after_chain_consolidation() -> eyre::Res
     fixture.l1().finalize_l1_block(batch_1_finalization_block_info.number).await?;
 
     // Wait for L1 block finalized event
-    fixture.expect_event().l1_block_finalized().await?;
+    fixture.expect_event().l1_genesis_finalized().await?;
 
     let status = fixture.get_sequencer_status().await?;
 
--- a/crates/primitives/src/batch.rs
+++ b/crates/primitives/src/batch.rs
@@ -1,4 +1,4 @@
-use crate::RollupNodePrimitiveParsingError;
+use crate::{BlockInfo, RollupNodePrimitiveParsingError};
 
 use super::L2BlockInfoWithL1Messages;
 
@@ -167,6 +167,12 @@ impl BatchConsolidationOutcome {
     /// Adds the skipped L1 messages indexes.
     pub fn with_skipped_l1_messages(&mut self, skipped: Vec<u64>) {
         self.skipped_l1_messages = skipped;
+    }
+
+    /// Returns the updated L2 head block info if consolidation updated it.
+    pub fn updated_l2_head(&self) -> Option<BlockInfo> {
+        self.l2_head_updated
+            .then(|| self.blocks.last().expect("must have at least one block").block_info)
     }
 }
 


yiweichi:
Hey @frisitano, I think we could have a first round of overview review. There are something still missing:
1. some tests won't pass, still checking the root cause.
2. need to add some database check
3. missing test for node shuts down scenario.
4. a lot of test logic could be improved

crates/node/src/test_utils/fixture.rs
```diff
@@ -201,6 +215,82 @@ impl TestFixture {
     ) -> eyre::Result<rollup_node_chain_orchestrator::ChainOrchestratorStatus> {
         self.get_status(0).await
     }
+
+    /// Get the Anvil HTTP endpoint if Anvil was started.
+    pub fn anvil_endpoint(&self) -> Option<String> {
+        self.anvil.as_ref().map(|a| a.http_endpoint())
+    }
```
frisitano: I think this should be replaced with `anvil_provider(&self) -> Option<RetryProvider>`

You can get this from `self.anvil.as_ref().map(|a| a.http_provider())`

crates/node/src/test_utils/fixture.rs
```diff
@@ -201,6 +215,82 @@ impl TestFixture {
...
+            self.anvil_endpoint().ok_or_else(|| eyre::eyre!("Anvil is not running"))?;
+
+        // Create RPC client
+        let client = alloy_rpc_client::RpcClient::new_http(anvil_endpoint.parse()?);
+
+        // Call anvil_reorg
+        // Parameters: (depth, transactions)
+        // - depth: number of blocks to rewind from current head
+        // - transactions: empty array means reorg without adding new transactions
+        let _: () = client.request("anvil_reorg", (depth, Vec::<String>::new())).await?;
```
frisitano: use `AnvilApi` as mentioned above.

crates/node/src/test_utils/fixture.rs
```diff
@@ -421,11 +520,63 @@ impl TestFixtureBuilder {
...
+    pub const fn with_anvil_chain_id(mut self, chain_id: u64) -> Self {
+        self.anvil_config.chain_id = Some(chain_id);
+        self
+    }
+
+    /// Set the block time for Anvil (in seconds).
+    pub const fn with_anvil_block_time(mut self, block_time: u64) -> Self {
+        self.anvil_config.block_time = Some(block_time);
+        self
+    }
```
frisitano: How about combining these all into one as:

```rust
with_anvil(&self, state_path: Optional<impl Into<PathBuf>, chain_id: Optional<u64> ...)
```

crates/providers/src/l1/blob/anvil.rs
```diff
@@ -21,11 +21,23 @@ impl AnvilBlobProvider {
...
 impl BlobProvider for AnvilBlobProvider {
+    #[allow(clippy::large_stack_frames)]
     async fn blob(
         &self,
         _block_timestamp: u64,
         hash: B256,
     ) -> Result<Option<Arc<Blob>>, L1ProviderError> {
-        Ok(self.inner.request("anvil_getBlobByHash", (hash,)).await?)
+        // Request blob data from Anvil - it returns hex-encoded bytes
+        let blob_data: Option<Bytes> = self.inner.request("anvil_getBlobByHash", (hash,)).await?;
```
frisitano: use `AnvilApi`

crates/watcher/src/lib.rs
```diff
@@ -99,6 +99,8 @@ pub struct L1Watcher<EP> {
     is_synced: bool,
     /// The log query block range.
     log_query_block_range: u64,
+    /// Test mode: skip sending `L1Notification::Synced` events.
+    test_mode_skip_synced_notification: bool,
```
frisitano: Can we put this behind a `test-utils` feature flag so that we only enable this in test mode?

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,566 @@
...
+        .followers(1)
+        .skip_l1_synced_notifications() // Prevents automatic L1Synced, simulates initial sync
+        .with_anvil()
+        .with_anvil_chain_id(22222222)
+        .build()
+        .await?;
+
+    // Record initial state - should be at genesis
+    let initial_status = fixture.get_sequencer_status().await?;
+    let initial_safe = initial_status.l2.fcs.safe_block_info().number;
```
frisitano: Should we assert `ininital_safe == 0`?

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,566 @@
...
+
+    // Decode hex string to bytes
+    let raw_tx_bytes = if let Some(stripped) = raw_tx_hex.strip_prefix("0x") {
+        alloy_primitives::hex::decode(stripped)?
+    } else {
+        alloy_primitives::hex::decode(raw_tx_hex)?
+    };
+
+    Ok(Bytes::from(raw_tx_bytes))
+}
```
frisitano: It seems wasteful that we read from file for every transaction. Can we read all test transactions into memory at once at the start of these tests?

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,579 @@
...
+    // Step 2: Send BatchCommit transactions to L1 while node is syncing
+    // These commits contain L2 blocks that should eventually become the safe head
+    for i in 0..=6 {
+        let commit_batch_tx = read_test_transaction("commitBatch", &i.to_string())?;
+        fixture.anvil_inject_tx(commit_batch_tx).await?;
+    }
+    // fixture.anvil_mine_blocks(1).await?;
+
+    // Allow time for L1 block processing
+    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
jonastheis: I think it would be better (faster test execution and less prone to local env) to either explicitly produce the block or wait for a block to be produced by polling every 50ms or something and then failing if no block was produced after a few seconds.

jonastheis: Alternatively (and maybe better) we can assert here the exact events that are expected to be indexed by the node. Here specifically it should be `BatchCommitIndexed` for each of the batches.

crates/node/src/test_utils/fixture.rs
```diff
@@ -201,6 +215,82 @@ impl TestFixture {
...
+        // Ensure Anvil is running
+        let anvil_endpoint =
+            self.anvil_endpoint().ok_or_else(|| eyre::eyre!("Anvil is not running"))?;
+
+        // Create RPC client
+        let client = alloy_rpc_client::RpcClient::new_http(anvil_endpoint.parse()?);
+
+        // Mine blocks using anvil_mine RPC method
+        // Parameters: (num_blocks, interval_in_seconds)
+        let _: () = client.request("anvil_mine", (num_blocks, 0)).await?;
```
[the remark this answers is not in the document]
yiweichi: Good idea, done in https://github.com/scroll-tech/rollup-node/commit/29dee411c81e30442e28b2960468cb7078ad5359

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,579 @@
...
+    let initial_status = fixture.get_status(0).await?;
+    let initial_safe = initial_status.l2.fcs.safe_block_info().number;
+    assert_eq!(initial_safe, 0, "Initial safe head should be at genesis (block 0)");
+
+    // Step 2: Send BatchCommit transactions (batches 0-6) to L1
+    for i in 0..=6 {
+        let commit_batch_tx = read_test_transaction("commitBatch", &i.to_string())?;
+        fixture.anvil_inject_tx(commit_batch_tx).await?;
+    }
+
```
jonastheis: should validate `BatchCommitIndexed`

crates/watcher/src/lib.rs
yiweichi: Sure, done in commit https://github.com/scroll-tech/rollup-node/pull/428/commits/731ae479f6d7843d0d3dd5648df91d8961a02532

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,603 @@
...
+///
+/// # Expected Behavior
+/// The finalized head should NOT change when `BatchFinalized` events are reorged.
+///
+/// This is by design: The node only updates its finalized head when a `BatchFinalized`
+/// event is **itself finalized on L1** (i.e., included in a finalized L1 block).
+/// Since L1 finality is irreversible (at least under honest majority assumption),
+/// a properly processed `BatchFinalized` event can never actually be reorged in practice.
+///
+/// This test simulates the impossible scenario to verify the node handles it gracefully
```
jonastheis: This line of the comment is wrong. The scenario is possible

crates/node/tests/l1_sync.rs
```diff
@@ -0,0 +1,603 @@
...
+        fixture.expect_event().batch_consolidated().await?;
+    }
+
+    // Step 3: Send BatchFinalized transactions (batches 1-2)
+    for i in 1..=2 {
+        let finalize_batch_tx = read_test_transaction("finalizeBatch", &i.to_string())?;
+        fixture.anvil_inject_tx(finalize_batch_tx).await?;
+    }
+    for _ in 1..=2 {
+        fixture.expect_event().batch_finalized().await?;
```
jonastheis: Why are the batches finalized here? They should only become finalized if the block carrying the finalized event also becomes finalized. imo we should avoid finalization of L1 here

crates/node/tests/l1_sync.rs
yiweichi: Could you elaborate it?

crates/chain-orchestrator/src/error.rs
```diff
@@ -92,6 +92,10 @@ pub enum ChainOrchestratorError {
     /// An error occurred while handling rollup node primitives.
     #[error("An error occurred while handling rollup node primitives: {0}")]
     RollupNodePrimitiveError(rollup_node_primitives::RollupNodePrimitiveError),
+    /// Shutdown requested - this is not a real error but used to signal graceful shutdown.
+    #[cfg(feature = "test-utils")]
+    #[error("Shutdown requested")]
+    Shutdown,
```
frisitano: I think we should change this to an event instead, `ChainOrchestratorEvent::Shutdown`

.github/workflows/test.yaml
```diff
@@ -51,9 +51,32 @@ jobs:
...
+        with:
+          cache-on-failure: true
+      - uses: taiki-e/install-action@nextest
+      - name: Run L1 sync integration tests (serial)
+        run: |
+          cargo nextest run --all-features --workspace --locked \
+            --no-tests=pass -E 'binary(l1_sync)' --test-threads=1 \
+            --failure-output immediate \
+            --success-output never
+
```
frisitano: Lets remove this.

crates/node/src/test_utils/reboot.rs
```diff
@@ -0,0 +1,241 @@
...
+    ///
+    /// // Shutdown follower node at index 1
+    /// fixture.shutdown_node(1).await?;
+    ///
+    /// // Node slot is now None, but indices are preserved
+    /// assert!(fixture.nodes[1].is_none());
+    ///
+    /// // Later, restart the node
+    /// fixture.start_node(1).await?;
+    /// ```
```
frisitano: update and make comment more concise.

crates/node/src/args.rs
```diff
@@ -465,7 +473,8 @@ impl ScrollRollupNodeConfig {
         )
         .await;
 
-        let (chain_orchestrator, handle) = ChainOrchestrator::new(
+        #[cfg_attr(not(feature = "test-utils"), allow(unused_mut))]
+        let (chain_orchestrator, mut handle) = ChainOrchestrator::new(
```
frisitano: combine this suggestion with the one below.
```suggestion
let (chain_orchestrator, handle) = ChainOrchestrator::new(
```

PR #456, opened 2025-12-07
fix: codec v7 version field

shu-unifra:


yiweichi:
Closed by this https://github.com/scroll-tech/rollup-node/pull/459

PR #457, opened 2025-12-09
fix: building in zkvm guest

lightsing:
dogeos using codec crate in guest, zstd won't compile on such target.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -117,7 +117,7 @@ jobs:
         with:
           cache-on-failure: true
       - name: Run feature check
-        run: cargo hack check --feature-powerset
+        run: cargo hack check --feature-powerset --mutually-exclusive-features "zstd,ruzstd" --at-least-one-of "zstd,ruzstd"
 
   no_std:
     runs-on: ubuntu-latest


crates/codec/src/decoding/v2/zstd.rs
```diff
@@ -2,14 +2,18 @@
 
 use std::io::Read;
 
-use zstd::Decoder;
+#[cfg(feature = "zstd")]
+use ruzstd as _;
```
Thegaram: `ruzstd` is only used for decoding? I guess it should be fully compatible with the C impl, right?

crates/codec/src/decoding/v2/zstd.rs
Thegaram: > zstd won't compile on such target

Why is that?

lispc: c code

Thegaram: We cannot use C compiled to RISC-V in the guest program?

crates/codec/src/decoding/v2/zstd.rs
Thegaram: > 2. gate v2::zstd, and make zstd as optional dep.

Does that mean that they don't actually need `decompress_blob_data`? Otherwise, without zstd, how would it work?

crates/codec/src/decoding/v2/zstd.rs
```diff
@@ -31,3 +35,20 @@ pub fn decompress_blob_data(data: &[u8]) -> Vec<u8> {
...
+    use ruzstd::decoding::StreamingDecoder;
+
+    let mut header_data = ZSTD_MAGIC_NUMBER.to_vec();
+    header_data.extend_from_slice(data);
+
+    // init decoder and owned output data.
+    let mut decoder = StreamingDecoder::new(header_data.as_slice()).unwrap();
+    // heuristic: use data length as the allocated output capacity.
+    let mut output = Vec::with_capacity(header_data.len());
+    decoder.read_to_end(&mut output).unwrap();
```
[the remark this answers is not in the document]
frisitano: I agree that is fine to update the API. The reason we used unwrap is that it is undefined behaviour in the protocol spec to have invalid data in batches. However, I agree that an update to the API is preferable.

lightsing:
@frisitano as you requested, but this brokes `--all-features`, what's your opinion on this? shall we update the ci or allow enable all features? cc @Thegaram

PR #484, opened 2026-01-28
introduce remote block source add on

frisitano:
This PR introduces an addon that allows a node to follow a remote node via rpc, import new blocks from the remote node and then build blocks on them.


yiweichi (approved):
lgtm

